PR #33, opened 2022-12-03
Power is derivative

cslade1970:


symplyphysics/laws/electricity/power_from_energy_time.py
```diff
@@ -0,0 +1,24 @@
...
+
+# Description
+# Power directly proportional to energy (work) and inversely proportional to time
+# P = Q / t
+# where:
+# Q  - some energy (work)
+# t -  energy action time
+
+power, energy, time = symbols('power energy time')
+law = Eq(power, energy / time)
```
blackyblack, line 14: Add TODO that this law should be derived from definition.

blackyblack (approved):

PR #66, opened 2023-04-13
internal_reflection

paradoxx84:


examples/optics/total_internal_reflection.py
```diff
@@ -0,0 +1,34 @@
...
+from sympy.physics.units.definitions.dimension_definitions import angle as angle_type
+
+# This example show usefulness of refraction law.
+## If ray of light tries to transfer from one medium to another, part of it transfers and refracts and another part reflects.
+## There is a possibility to have the whole ray to be reflected and nothing refracted to another medium.
+## It is possible when ray transfers from more denser medium to less denser.
+## This phenomenon is utilized by fiber optics technology.
+
+## As law says, if alpha is accidence angle, beta is refraction angle, n1*sin(alpha) = n2*sin(beta), or sin(beta) / sin(alpha) = n1/n2. If n1 > n2, beta will be always more then alpha.
+## If we enlarge aplpha, beta enlarges as well. There is such value of alpha when beta reaches it's max possible value - 90grads, it means refracted ray goes along mediums border.
```
blackyblack, line 16:
```suggestion
## If we increase alpha, beta increases as well. There is such value of alpha when beta reaches it's max possible value - 90 degrees, it means refracted ray goes along mediums border.
```

examples/optics/total_internal_reflection.py
```diff
@@ -0,0 +1,34 @@
...
+## If we continue enlarging alpha, there will be no more refraction anymore, the whole ray will be only reflected.
```
blackyblack, line 17:
```suggestion
## If we continue increasing alpha, there will be no more refraction anymore, the whole ray will be only reflected.
```

examples/optics/total_internal_reflection.py
```diff
@@ -0,0 +1,34 @@
...
+coat_refractive_index = 1.474
+
+
+extreme_angle = incedence_angle.subs({refraction_law.incedence_refractive_index: core_refractive_index,
+                                      refraction_law.resulting_refractive_index: coat_refractive_index,
+                                      refraction_law.refraction_angle: pi/2
+                                    })
+
+
+print(f"{extreme_angle}")
```
blackyblack, line 33:
```suggestion
print(f"Incedence angle with full reflection: {extreme_angle} radian")
```

blackyblack, line 33: Printing result in degrees should be better.

blackyblack (approved):

PR #81, opened 2023-06-14
Energy conservation law

blackyblack:


symplyphysics/laws/conservation/mechanical_energy_after_equals_to_mechanical_energy_before.py
```diff
@@ -0,0 +1,48 @@
+from sympy import (Eq, solve, dsolve)
+from symplyphysics import (units, expr_to_quantity, Quantity, Symbol, print_expression, Function,
+    validate_input, validate_output)
+from symplyphysics.core.expr_comparisons import expr_equals
+from symplyphysics.laws.conservation import mechanical_energy_is_constant
+
+# Description
+## The total mechanical energy of a system is conserved i.e., the energy can neither be created nor be destroyed.
```
paradoxx84, line 8: Isolated system

symplyphysics/laws/conservation/mechanical_energy_after_equals_to_mechanical_energy_before.py
```diff
@@ -0,0 +1,48 @@
...
+## It can only be internally converted from one form to another if the forces doing work on the system are conservative in nature.
```
paradoxx84, line 9: If work (of external forces) is applied to system, system energy changes.

paradoxx84 (changes requested):

symplyphysics/laws/conservation/mechanical_energy_after_equals_to_mechanical_energy_before.py
blackyblack: It should not change when only conservative forces are applied. For example, falling due to gravity object loses potential energy and gain kinetic energy. Here external gravitational force is applied but total energy does not change.

paradoxx84 (approved):

PR #132, opened 2023-12-26
The_Law_of_Refraction_of_Light

lrtrtknt:


The_Law_of_Refraction_of_Light.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+from sympy import sin, rad
+from your_optics_module import calculate_refraction_angle
```
blackyblack, line 3: There is no such module.

The_Law_of_Refraction_of_Light.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+from sympy import sin, rad
+from your_optics_module import calculate_refraction_angle
+
+# Replace 'your_optics_module' with the actual name of your code file
```
blackyblack, line 5: Don't try to game the test task by using ChatGPT.

blackyblack (changes requested):

PR #133, opened 2023-12-26
Test task.

molodora:
Add law of photoelectric effect and test for it.


test/waves/photoelectric_effect.py
```diff
@@ -0,0 +1,46 @@
+from collections import namedtuple
```
blackyblack, line 1: Test's name should end with `test`

symplyphysics/laws/waves/photoelectric_effect.py
```diff
@@ -0,0 +1,37 @@
+from sympy import (Eq, solve)
```
blackyblack, line 1: We have a convention of not using marketing names, like "Newton's first law" or such. Write law name as "what_we_get_from_what_we_put" or "value_is_something" for definitions.

blackyblack (approved):

symplyphysics/laws/waves/photoelectric_effect.py
blackyblack: By the way, maybe this law should be in a `quantum` branch of laws?

test/waves/photoelectric_effect.py
molodora: Changed filename to `photoelectron_energy_from_frequency_test.py`

blackyblack (approved):

symplyphysics/laws/waves/photoelectric_effect.py
molodora: Changed filename to `photoelectron_energy_from_frequency.py`.

It seems to me, there're several files that should be in `quantum` folder, such as `photon_energy...` or `photon_momentum...`. Or may be it shoud be named `atomic`, for both particles and waves?

PR #157, opened 2024-01-05
Create law for distance from constant acceleration

Voldoned:
Motion with constant acceleration is described by the equation: $S = S_0 + V_0 t + at^2/2$,
where $S_0$ is the distance in the start moment, $V_0$ is the velocity in the start moment, $a$ is the acceleration, $t$ is time.

### DIFF
--- a/symplyphysics/laws/kinematic/distance_from_constant_acceleration.py
+++ b/symplyphysics/laws/kinematic/distance_from_constant_acceleration.py
@@ -1,0 +1,36 @@
+from sympy import (Eq, solve)
+from symplyphysics import (units, Quantity, Symbol, print_expression, validate_input,
+    validate_output)
+
+# Description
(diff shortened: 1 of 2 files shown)


symplyphysics/laws/kinematic/distance_from_constant_acceleration.py
```diff
@@ -0,0 +1,36 @@
+from sympy import (Eq, solve)
+from symplyphysics import (units, Quantity, Symbol, print_expression, validate_input,
+    validate_output)
+
+# Description
+## Motion with constant acceleration is described by the equation: S = S0 + V0 * t + (a * t^2)/2
```
blackyblack, line 6: We already have this: https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/kinematic/constant_acceleration_movement_is_parabolic.py

blackyblack (changes requested):

PR #158, opened 2024-01-05
Create law for impuls

Voldoned:
The momentum of a body (the amount of motion) is a vector physical quantity, the modulus of which is equal to the product of the mass of the body by its speed:
$p = m * v$,
where $m$ - mass of the body, $v$ - velocity.

### DIFF
--- a/symplyphysics/laws/kinematic/impuls.py
+++ b/symplyphysics/laws/kinematic/impuls.py
@@ -23,8 +23,8 @@ def print_law() -> str:
 
 
 @validate_input(mass_of_the_body_=mass_of_the_body, velocity_=velocity)
-@validate_output(velocity)
+@validate_output(impuls)
 def calculate_impuls(mass_of_the_body_: Quantity, velocity_: Quantity) -> Quantity:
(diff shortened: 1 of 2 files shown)


symplyphysics/laws/kinematic/impuls.py
```diff
@@ -0,0 +1,30 @@
+from sympy import (Eq, solve)
+from symplyphysics import (units, Quantity, Symbol, print_expression,
+                           validate_input, validate_output)
+
+
+# Descroption
+## The momentum of a body (the amount of motion) is a vector physical quantity, the modulus of which is equal to the product of the mass of the body by its speed.
+## p = m * v
```
blackyblack, line 8: We already have this: https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/definitions/momentum_is_mass_times_velocity.py

blackyblack (changes requested):

symplyphysics/laws/kinematic/impuls.py
Voldoned: I'm sorry.

PR #175, opened 2024-01-09
Parallel resistors (redone)

alesanter:


symplyphysics/laws/electricity/circuits/resistivity_of_parallel_resistors.py
```diff
@@ -0,0 +1,36 @@
...
+law = Eq(parallel_resistance, 1 / SumArray(inverse_resistances), evaluate=False)
+
+
+def print_law() -> str:
+    return print_expression(law)
+
+
+@validate_input(inverse_resistances_=inverse_resistances)
+@validate_output(units.impedance)
+def calculate_parallel_resistance(inverse_resistances_: list[Quantity]) -> Quantity:
```
blackyblack, line 30: Accept direct resistance here. There is no sense in using inverse resistance as parameter.

blackyblack (approved):

symplyphysics/laws/electricity/circuits/resistivity_of_parallel_resistors.py
alesanter: I understand that, but that would require the law to contain direct resistances too, and I'm not sure what change to make in the law so that it contains a sum of inverted resistances.

blackyblack: No, it does not require to change law. Give it a try and ask me for help if you're stuck.

alesanter: I tried to implement a workaround but it seems to require more changes to files (so that mypy doesn't fail) and probably there's a better way to do that

PR #180, opened 2024-01-10
Create example for parallel and serial connection of resistors

alesanter:


examples/electricity/two_resistors.py
```diff
@@ -0,0 +1,65 @@
...
+from symplyphysics.laws.electricity import current_is_proportional_to_voltage as ohm_law
+from symplyphysics.laws.electricity.circuits import resistivity_of_parallel_resistors as parallel_resistors
+from symplyphysics.laws.electricity.circuits import resistivity_of_serial_resistors as serial_resistors
+
+# Two resistors are connected across a 12 V battery with internal resistance of 1 Ohm.
+# When they are connected in parallel, the current in the circuit is 4 A.
+# When they are connected in series, the current in the circuit is 1 A.
+# Find the resistance of said resistors.
+
+R1 = Symbol("R1", units.impedance)
```
blackyblack, line 14: Use sympy symbols and remove dimensions.

examples/electricity/two_resistors.py
```diff
@@ -0,0 +1,65 @@
...
+R2 = Symbol("R2", units.impedance)
+R_battery = Symbol("Rb", units.impedance)
+E_battery = Symbol("E", units.voltage)
+I_series = Symbol("Is", units.current)
+I_parallel = Symbol("Ip", units.current)
+
+# Parallel connection
+
+parallel_law = parallel_resistors.law.subs({parallel_resistors.inv_resistances: (1 / R1, 1 / R2)})
```
blackyblack, line 23: Use conductivity law to find inverted resistance: https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/definitions/electrical_conductivity_is_inversed_resistance.py
Also, let's use `conductance` everywhere instead of `inv_resistance`. For this reason, the law for parallel resistors should be updated as well.

blackyblack (approved):

examples/electricity/two_resistors.py
alesanter: Okay, other examples use `Symbol` though.

examples/electricity/two_resistors.py
```diff
@@ -0,0 +1,80 @@
...
+R_battery = symbols("R_battery")
+E_battery = symbols("E_battery")
+I_series = symbols("I_series")
+I_parallel = symbols("I_parallel")
+
+# Parallel connection
+
+## Find resistance R12 using the law of conductance
+
+sigma1 = 1 / R1
```
blackyblack, line 26: Calculate using the corresponding law.

blackyblack (changes requested):

examples/electricity/two_resistors.py
blackyblack: There are two different `Symbol`s. One from sympy and another from symplyphysics. Use the sympy's one.

blackyblack (approved):

PR #182, opened 2024-01-10
Add law and test for equation of continuity (fluid mechanics)

alesanter:


symplyphysics/laws/hydro/volume_flux_is_constant.py
```diff
@@ -0,0 +1,40 @@
...
+    Symbol,
+    validate_input,
+    validate_output,
+)
+
+# Description
+## The product of the area and the fluid speed, which is called volume flux, is constant 
+## at all points along the tube of flow of an incompressible liquid.
+
+# Law: A*v = constant
```
blackyblack, line 14: We usually write constant bound laws as derivative equals zero, eg https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/conservation/mechanical_energy_is_constant.py
Maybe write an additional form of this law? Is there a derivative version of this law for reference?

symplyphysics/laws/hydro/volume_flux_is_constant.py
```diff
@@ -0,0 +1,40 @@
...
+## A - area of the tube of flow
+## v - fluid speed
+
+# It is also called the equation of continuity
+
+# Condition
+## The fluid should be ideal, i.e. nonviscous, in laminar flow, incompressible and irrotational.
```
blackyblack, line 21: Separate conditions.

symplyphysics/laws/hydro/volume_flux_is_constant.py
```diff
@@ -0,0 +1,40 @@
...
+
+area_start = Symbol("A_start", units.area)
+v_start = Symbol("v_start", units.velocity)
```
blackyblack, line 24: Don't shortcut. Use velocity instead of `v`

blackyblack (approved):

symplyphysics/laws/hydro/volume_flux_is_constant.py
blackyblack: Also instead of area_start, area_end we can use a bit more pleasant area(time_before), like here: https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/conservation/momentum_after_collision_equals_to_momentum_before.py

alesanter: Okay, I took inspiration from https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/thermodynamics/volume_is_constant.py and the like.

blackyblack: I see. The preferred way depends, on whether area and velocity are non-linear functions. If they usually are, it is better to have a derivative.

symplyphysics/laws/hydro/volume_flux_is_constant.py
```diff
@@ -0,0 +1,43 @@
...
+
+law = Eq(Derivative(tube_area(time) * fluid_speed(time), time), 0)
+
+
+@validate_input(tube_area_before_=tube_area, fluid_speed_before_=fluid_speed, tube_area_after_=tube_area)
+@validate_output(units.velocity)
+def calculate_flow_speed(tube_area_before_: Quantity, fluid_speed_before_: Quantity, tube_area_after_: Quantity) -> Quantity:
+    solved = dsolve(law, fluid_speed(time))
+    result_expr = solved.subs({
+        "C1": tube_area_before_ * fluid_speed_before_,
```
blackyblack, line 40: We should explain, where it came from. I suggest:
1. Find solution for differential equation: `dsolved = dsolve(law, fluid_speed(time))`
2. C1 is the same for time_before and time_after.
3. Find C1: `c1_value = solve(dsolved, "C1")[0].subs({tube_area(time): tube_area_before, fluid_speed(time): fluid_speed_before})`
4. Find resulting velocity: `result_expr = dsolved.subs({"C1": c1_value, tube_area(time): tube_area_after_})`

blackyblack (approved):

PR #191, opened 2024-01-11
Derive proof of equivalence for laws of parallel resistors

alesanter:


symplyphysics/laws/electricity/circuits/resistivity_of_parallel_resistors.py
```diff
@@ -21,6 +25,57 @@
...
+parallel_resistance_from_conductivity_law = solve(
+    conductance_definition.definition.subs(
+        conductance_definition.object_conductivity,
+        parallel_conductance
+    ),
+    conductance_definition.object_resistance
+)[0]
+
+parallel_resistance_from_law_in_question = solve(
+    law.subs(inv_resistances, (1 / resistance1, 1 / resistance2)),
```
blackyblack, line 69: We should not exlplictly divide, when proving something. Use conductance law instead.

symplyphysics/laws/electricity/circuits/resistivity_of_parallel_resistors.py
```diff
@@ -21,6 +25,57 @@
 law = Eq(parallel_resistance, 1 / SumArray(inv_resistances), evaluate=False)
```
blackyblack, line 25: I suggest to rename `inv_resistances` to `conductances`. This way we will justify the usage of conductance law in the proof. And also link together resistance and conductance laws.

blackyblack (approved):

PR #202, opened 2024-01-13
Law potential force is gradient

alesanter:
Unsure if there is a better way of writing the law and calculation, since I haven't found an analogue of `QuantityVector` for fields, but I tried my best.

### DIFF
--- a/symplyphysics/core/fields/operators.py
+++ b/symplyphysics/core/fields/operators.py
@@ -7,7 +7,7 @@ from ..coordinate_systems.coordinate_systems import CoordinateSystem
 from ..vectors.vectors import Vector
 
 
-def gradient_operator(field: ScalarField) -> VectorField:
+def gradient_operator(field: ScalarField) -> Vector:
     field_space = field.apply_to_basis()
     if field.coordinate_system.coord_system_type == CoordinateSystem.System.CARTESIAN:
         x = field.coordinate_system.coord_system.base_scalars()[0]
@@ -16,9 +16,9 @@ def gradient_operator(field: ScalarField) -> VectorField:
         gradient = Vector([
             diff(field_space, x),
             diff(field_space, y),
-            diff(field_space, z)
+            diff(field_space, z),
         ], field.coordinate_system)
-        return VectorField.from_vector(gradient)
+        return gradient
     if field.coordinate_system.coord_system_type == CoordinateSystem.System.CYLINDRICAL:
         r = field.coordinate_system.coord_system.base_scalars()[0]
         theta = field.coordinate_system.coord_system.base_scalars()[1]
@@ -26,19 +26,19 @@ def gradient_operator(field: ScalarField) -> VectorField:
         gradient = Vector([
             diff(field_space, r),
             diff(field_space, theta) / r,
-            diff(field_space, z)
+            diff(field_space, z),
         ], field.coordinate_system)
-        return VectorField.from_vector(gradient)
+        return gradient
     if field.coordinate_system.coord_system_type == CoordinateSystem.System.SPHERICAL:
         r = field.coordinate_system.coord_system.base_scalars()[0]
-        theta = field.coordinate_system.coord_system.base_scalars()[1]
-        phi = field.coordinate_system.coord_system.base_scalars()[2]
+        phi = field.coordinate_system.coord_system.base_scalars()[1]
+        theta = field.coordinate_system.coord_system.base_scalars()[2]
         gradient = Vector([
             diff(field_space, r),
+            diff(field_space, phi) / (r * sin(theta)),
             diff(field_space, theta) / r,
-            diff(field_space, phi) / (r * sin(theta))
         ], field.coordinate_system)
-        return VectorField.from_vector(gradient)
+        return gradient
     raise ValueError(f"Unsupported coordinate system: {field.coordinate_system}")
 
 
--- a/symplyphysics/core/fields/operators.py
+++ b/symplyphysics/core/fields/operators.py
@@ -31,12 +31,12 @@ def gradient_operator(field: ScalarField) -> Vector:
         return gradient
     if field.coordinate_system.coord_system_type == CoordinateSystem.System.SPHERICAL:
         r = field.coordinate_system.coord_system.base_scalars()[0]
-        phi = field.coordinate_system.coord_system.base_scalars()[1]
-        theta = field.coordinate_system.coord_system.base_scalars()[2]
+        theta = field.coordinate_system.coord_system.base_scalars()[1]
+        phi = field.coordinate_system.coord_system.base_scalars()[2]
         gradient = Vector([
             diff(field_space, r),
-            diff(field_space, phi) / (r * sin(theta)),
-            diff(field_space, theta) / r,
+            diff(field_space, theta) / (r * sin(phi)),
+            diff(field_space, phi) / r,
         ], field.coordinate_system)
         return gradient
     raise ValueError(f"Unsupported coordinate system: {field.coordinate_system}")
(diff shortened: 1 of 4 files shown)


symplyphysics/core/fields/operators.py
```diff
@@ -2,10 +2,46 @@
...
+        ], field.coordinate_system)
+        return VectorField.from_vector(gradient)
+    if field.coordinate_system.coord_system_type == CoordinateSystem.System.SPHERICAL:
+        r = field.coordinate_system.coord_system.base_scalars()[0]
+        theta = field.coordinate_system.coord_system.base_scalars()[1]
+        phi = field.coordinate_system.coord_system.base_scalars()[2]
+        gradient = Vector([
+            diff(field_space, r),
+            diff(field_space, theta) / r,
+            diff(field_space, phi) / (r * sin(theta))
```
blackyblack, line 39: This is wrong. Azimuthal and polar angles have different notation here than in some books.

symplyphysics/core/fields/operators.py
```diff
@@ -2,10 +2,46 @@
 
 from ..dimensions import ScalarValue
 from ..fields.vector_field import VectorField
+from ..fields.scalar_field import ScalarField
 from ..coordinate_systems.coordinate_systems import CoordinateSystem
 from ..vectors.vectors import Vector
 
 
+def gradient_operator(field: ScalarField) -> VectorField:
```
blackyblack, line 10: It is more usual to have Vector as a result of gradient operation.

symplyphysics/laws/dynamics/fields/conservative_force_is_gradient_of_potential_energy.py
```diff
@@ -0,0 +1,25 @@
...
+## For every conservative force there exists a scalar field (potential), for which applies
+## the following equality:
+
+# Law: F = - grad U
+## F - vector field of the conservative force
+## U - scalar field of its potential
+## grad - gradient operator
+
+def law(potential_: ScalarField) -> VectorField:
+    gradient_field = gradient_operator(potential_).apply_to_basis()
```
blackyblack, line 20: If we get Vector from gradient, this code would be easier.

test/core/fields/operators_test.py
```diff
@@ -24,6 +25,53 @@ def test_args_fixture():
...
+    result_vector = result_field.apply_to_basis()
+    x = field.coordinate_system.coord_system.base_scalars()[0]
+    y = field.coordinate_system.coord_system.base_scalars()[1]
+    z = field.coordinate_system.coord_system.base_scalars()[2]
+    for i, expr in enumerate([2 * x, 2 * y, - 2 * z]):
+        assert expr_equals(result_vector.components[i], expr)
+
+
+def test_cylindrical_gradient(test_args):
+    # verify that in cylindrical coordinates the result is same
```
blackyblack, line 42: The same with what?

test/core/fields/operators_test.py
```diff
@@ -24,6 +25,53 @@ def test_args_fixture():
...
+    C1 = CoordinateSystem(CoordinateSystem.System.CYLINDRICAL)
+
+    def field_function(p: CylinderPoint) -> ScalarValue:
+        return p.r ** 2 - p.z ** 2
```
blackyblack, line 46: Theta is not tested.

test/core/fields/operators_test.py
```diff
@@ -24,6 +25,53 @@ def test_args_fixture():
...
+    result_cylindrical_vector = result_cylindrical_field.apply_to_basis()
+
+    r = cylindrical_field.coordinate_system.coord_system.base_scalars()[0]
+    z = cylindrical_field.coordinate_system.coord_system.base_scalars()[2]
+    for i, expr in enumerate([2 * r, 0, - 2 * z]):
+        assert expr_equals(result_cylindrical_vector.components[i], expr)
+
+
+def test_spherical_gradient(test_args):
+    # verify that in spherical coordinates the result is the same
```
blackyblack, line 59: Remove

test/core/fields/operators_test.py
```diff
@@ -24,6 +25,53 @@ def test_args_fixture():
...
+    C1 = CoordinateSystem(CoordinateSystem.System.SPHERICAL)
+
+    def field_function(p: SpherePoint) -> ScalarValue:
+        return - p.r**2 * cos(2 * p.theta)
```
blackyblack, line 63: Phi is not tested

test/core/fields/operators_test.py
```diff
@@ -24,6 +25,53 @@ def test_args_fixture():
...
+
+    spherical_field = ScalarField(field_function, C1)
+    result_spherical_field = gradient_operator(spherical_field)
+    result_spherical_vector = result_spherical_field.apply_to_basis()
+
+    r = spherical_field.coordinate_system.coord_system.base_scalars()[0]
+    theta = spherical_field.coordinate_system.coord_system.base_scalars()[1]
+    for i, expr in enumerate([-2 * r * cos(2 * theta), 2 * r * sin(2 * theta), 0]):
```
blackyblack, line 71: Where did you get the test value? How come it works despite the gradient operator is wrong?

test/dynamics/fields/conservative_force_is_gradient_of_potential_energy_test.py
```diff
@@ -0,0 +1,38 @@
...
+    result_vector = result_field.apply_to_basis()
+
+    x = result_field.coordinate_system.coord_system.base_scalars()[0]
+    for i, expr in enumerate([-x, 0, 0]):
+        assert expr_equals(result_vector.components[i], expr)
+
+
+def test_bad_law(test_args):
+    bad_field = VectorField(lambda point: [point.x, point.y, point.z])
+    with raises(Exception):
```
blackyblack, line 35: What exception exactly?

symplyphysics/core/fields/operators.py
alesanter: Isn't it better to have `VectorField` in case the end user has to plug the result of `gradient_operator` into other functions that accept `VectorField` instead of `Vector`, like `divergence_operator`? And as far as I understand, to get a `Vector` from `VectorField` it's as easy as to call `.apply_to_basis()` on the latter.

symplyphysics/core/fields/operators.py
alesanter: I now understand that in your notation the angle names are swapped. However, it most literature and scientific journals  [like sciencedirect](https://www.sciencedirect.com/topics/mathematics/spherical-polar-coordinate) other notation is accepted, namely theta is the polar angle and phi is the azimuthal one. Maybe it is better to stick to more a more common convention so as not to make confusion?

blackyblack: You're right but it is a major refactor and out of the scope of this PR

symplyphysics/core/fields/operators.py
blackyblack: Sometimes it results with plain scalar values and it does not make sense to make a vector field from them. I believe making it return Vector would make code cleaner.

alesanter: I agree with you now

symplyphysics/laws/dynamics/fields/conservative_force_is_gradient_of_potential_energy.py
```diff
@@ -0,0 +1,22 @@
...
+from symplyphysics.core.fields.scalar_field import ScalarField
+from symplyphysics.core.fields.vector_field import VectorField
+from symplyphysics.core.fields.operators import gradient_operator
+
+# Description
+## A conservative force is a such a force, the total work of which in moving a particle
+## between two points is independent of the path taken. Alternative definition states that
+## if a particle travels in a closed loop, the total work done by a conservative force is zero.
+
+## For every conservative force there exists a scalar field (potential), for which applies
```
blackyblack, line 11: Add to conditions, that force is conservative.

blackyblack (approved):

PR #212, opened 2024-01-14
Create law and test: volume flow rate

Voldoned:


symplyphysics/definitions/volume_flow_rate.py
```diff
@@ -0,0 +1,63 @@
...
+})
+volume_equation = dsolve(volume_rate_equation, volume(time)).subs({
+    "C1": density_law.volume
+})
+
+density_equation = density_law.definition.subs({
+    density_law.mass: mass_equation.rhs,
+    density_law.volume: volume_equation.rhs
+})
+assert expr_equals(density_law.definition.rhs, density_equation.rhs)
```
blackyblack, line 47: Definitions do not require proofs.

symplyphysics/definitions/volume_flow_rate.py
```diff
@@ -0,0 +1,40 @@
+from sympy import (Eq, Derivative, dsolve)
+from symplyphysics import (units, Quantity, Function, Symbol, print_expression, validate_input,
+    validate_output)
+from symplyphysics.core.expr_comparisons import expr_equals
+from symplyphysics.definitions import mass_flow_rate as mass_flow_rate_law
+from symplyphysics.definitions import density_from_mass_volume as density_law
+
+# Description
+## The rate of change in the volume of matter. For example, the outflow of a substance
```
blackyblack, line 9: Do we really need this law? Is there any case when we cannot use mass flow rate?

PR #216, opened 2024-01-15
Create definition and test: mass fraction of solute

Voldoned:


symplyphysics/definitions/mass_fraction_of_the_solute.py
```diff
@@ -0,0 +1,35 @@
+from sympy import (Eq, solve)
+from symplyphysics import (units, Quantity, Symbol, print_expression, validate_input,
+    validate_output, dimensionless)
+
+# Description
+## The mass fraction is the ratio of the mass of the solute to the total mass of the solution
```
blackyblack, line 6: I do not like the word `solute` here. Check for more common definition of the mass fraction.

symplyphysics/definitions/mass_fraction_of_the_solute.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def print_law() -> str:
+    return print_expression(definition)
+
+
+@validate_input(mass_of_component_=mass_of_component, mass_of_mixture_=mass_of_mixture)
+@validate_output(mass_fraction)
+def calculate_mass_fraction(mass_of_component_: Quantity,
+    mass_of_mixture_: Quantity) -> Quantity:
```
blackyblack, line 29: Return `float`

symplyphysics/definitions/mass_fraction_of_the_solute.py
```diff
@@ -0,0 +1,35 @@
...
+definition_units_SI = dimensionless
+
+
+def print_law() -> str:
+    return print_expression(definition)
+
+
+@validate_input(mass_of_component_=mass_of_component, mass_of_mixture_=mass_of_mixture)
+@validate_output(mass_fraction)
+def calculate_mass_fraction(mass_of_component_: Quantity,
```
blackyblack, line 28: I assume fraction is never higher than 100%. I suggest to create new `Fraction` type. You can use `Probability` as a reference.

symplyphysics/definitions/mass_fraction_of_the_solute.py
Voldoned: > Мне не нравится это слово `solute`здесь. 

And if I replace it with 'mixture'?

symplyphysics/definitions/mass_fraction_of_the_solute.py
Voldoned: Like this?

```python
from typing import Self


class Fraction(float):
    
    def __new__(cls, value: float) -> Self:
        if value < 0 or value > 1.0:
            raise AttributeError(f"Fraction value should be in range [0..1], got '{value}'")
        return float.__new__(cls, value)
```

symplyphysics/definitions/mass_fraction_of_the_solute.py
blackyblack: Mixture is synonim of `solution`. Did you mean `component`?

symplyphysics/definitions/mass_fraction_of_the_solute.py
blackyblack: Yes. It is copypaste though.

symplyphysics/definitions/mass_fraction_of_the_solute.py
Voldoned: I find basically a similar definition of this term:

> The mass fraction of a substance within a mixture is the ratio of the mass of that substance to the total mass of the mixture.

And I thought to correct the description as follows:
```python
# Description
## The mass fraction of a substance within a mixture is the ratio 
## of the mass of that substance to the total mass of the mixture:
## omega = m_i / m
## Where:
## m_i - mass of component
## m - mass of mixture
## omega - mass fraction of the mixture
```

symplyphysics/definitions/mass_fraction_of_the_solute.py
Voldoned: We can inherit the 'Probability' class or vice versa, inherit the 'Probability' class from this class. But since the description of the 'AttributeError' error is changing, we have to recreate the `__new__()` method. Alternatively, the check for belonging to the interval [0, 1] should be placed in a separate class method, and then reused, while generating different error descriptions.

blackyblack: For such trivial implementation it is better to copypaste and not worry about inheritance.

symplyphysics/definitions/mass_fraction_of_the_solute.py
blackyblack: Yes, it's better this way.

symplyphysics/definitions/mass_fraction_of_the_solute.py
Voldoned: Ok.

symplyphysics/core/symbols/fraction.py
```diff
@@ -0,0 +1,9 @@
+from typing import Self
+
+
+class Fraction(float):
+
+    def __new__(cls, value: float) -> Self:
+        if value < 0 or value > 1.0:
+            raise AttributeError(f"Fraction value should be in range [0..1], got '{value}'")
+        return float.__new__(cls, value)
```
blackyblack, line 9: Add new line

symplyphysics/definitions/mass_fraction.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def print_law() -> str:
+    return print_expression(definition)
+
+
+@validate_input(mass_of_component_=mass_of_component, mass_of_mixture_=mass_of_mixture)
+@validate_output(mass_fraction)
+def calculate_mass_fraction(mass_of_component_: Quantity,
+    mass_of_mixture_: Quantity) -> float:
```
blackyblack, line 29: What's the point of adding Fraction and not using it?
Return `Fraction`

test/definitions/mass_fraction_test.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_bad_mass_of_component(test_args):
+    mb = Quantity(1 * units.meter)
+    with raises(errors.UnitsError):
+        mass_fraction_law.calculate_mass_fraction(mb, test_args.m)
+    with raises(TypeError):
+        mass_fraction_law.calculate_mass_fraction(100, test_args.m)
+
+
+def test_bad_mass_of_solute(test_args):
```
blackyblack, line 32: We decided not to use `solute`

test/definitions/mass_fraction_test.py
```diff
@@ -0,0 +1,37 @@
...
+
+@fixture(name="test_args")
+def test_args_fixture():
+    m_i = Quantity(1 * units.kilogram)
+    m = Quantity(3 * units.kilogram)
+    Args = namedtuple("Args", ["m_i", "m"])
+    return Args(m_i=m_i, m=m)
+
+
+def test_basic_mass_fraction(test_args):
```
blackyblack, line 19: Now, when you have `Fraction`, could you test it? What if we try to calculate the fraction of 3 kg in 1 kg total mass?

PR #223, opened 2024-01-16
Charge conservation and quantization

alesanter:


symplyphysics/laws/electricity/charge_is_quantized.py
```diff
@@ -0,0 +1,35 @@
...
+
+# Law: q = n*e
+## q - electric charge
+## n - an integer factor
+## e - elementary charge
+
+charge = Symbol("charge", units.charge)
+integer_factor = Symbol("factor", dimensionless)
+
+law = Eq(charge, integer_factor * units.elementary_charge)
```
blackyblack, line 23: It assumes that any charge can only be negative (or positive, depending on elementary charge definition). Probably we should tell about charge magnitude and use absolute values.

symplyphysics/laws/conservation/charge_is_constant.py
```diff
@@ -0,0 +1,37 @@
...
+
+def print_law() -> str:
+    return print_expression(law)
+
+
+@validate_input(total_charge_before_=total_charge)
+@validate_output(total_charge)
+def calculate_charge_after(total_charge_before_: Quantity) -> Quantity:
+    dsolved = dsolve(law, total_charge(time))
+    result_charge = dsolved.subs("C1", total_charge_before_).rhs
```
blackyblack, line 36: Here you should know, what's on the left side of the result of dsolve(). I suggest not to rely on it. Instead you can take the dsolved, which tells that charge equals to `C1`. And solve a set of 2 equations: charge_before = C1, charge_after = C1. Solution will tell that charge_after = charge_before.

symplyphysics/laws/conservation/charge_is_constant.py
```diff
@@ -0,0 +1,37 @@
+from sympy import Eq, dsolve, Derivative
```
blackyblack, line 1: Next time separate features - each feature is a separate PR.

alesanter: Yes, had to learn it the hard way. Much better to have them separate.

symplyphysics/laws/conservation/charge_is_constant.py
alesanter: I see, it's just that I sometimes use already existing laws and they are written that way (namely, mass_is_constant and mechanical_energy_is_constant are the ones I referenced). I see the advantage of your suggestion, though. I just want to say that it would probably be easier if there were some conventions throughout the whole project which developers would follow.

symplyphysics/laws/electricity/charge_is_quantized.py
alesanter: As far as I know, most of the time in the English terminology the word `integer` refers to both positive and negative numbers. It would probably suffice to add a comment about that.

symplyphysics/laws/conservation/charge_is_constant.py
blackyblack: We cannot cover every line of code with conventions. It is better to add TODO hint everywhere you see the same problem.

symplyphysics/laws/electricity/charge_is_quantized.py
blackyblack: I do not think that proton has negative amount of charges.

symplyphysics/laws/conservation/charge_is_constant.py
alesanter: Okay, I see

symplyphysics/laws/electricity/charge_is_quantized.py
blackyblack: I think it should be signed. Can you give me a link to statement that elementary charge is always positive?

alesanter: In SI units it is defined as such: https://www.bipm.org/utils/common/pdf/si-brochure/SI-Brochure-9.pdf

image

Translating from French: ...It is defined taking the fixed numeric value of the elementary charge e equal to 1,602... ... This definition implies the exact relation e = 1,602...

blackyblack: Thanks.

blackyblack (approved):

PR #228, opened 2024-01-17
Create example for movement of charged particle in electric field

alesanter:


examples/electricity/charged_particle_in_electric_field.py
```diff
@@ -0,0 +1,73 @@
...
+## The plates are charged and produce a uniform, downward directed (y being the vertical axis) electric
+## field at all points between them. The electric field's magnitude is 1.1e6 N/C. What is the vertical
+## deflection of the drop at the far edge of the plates? Neglect the gravitational force on the drop
+## since the drop is small enough that the electrostatic force prevails over the gravitational one.
+
+drop_mass = Symbol("drop_mass", units.mass)
+drop_charge = Symbol("drop_charge", units.charge)
+drop_speed_x = Symbol("drop_speed_x", units.velocity)
+plate_length_x = Symbol("plate_length_x", units.length)
+electric_field_magnitude = Symbol("electric_field_magnitude", units.force / units.charge)
```
blackyblack, line 22: You can remove dimensions in examples. They are not verified hence easy to make a mistake in them.

examples/electricity/charged_particle_in_electric_field.py
```diff
@@ -0,0 +1,73 @@
...
+
+values = {
+    drop_mass: Quantity(1.5e-7 * units.gram),
+    drop_charge: Quantity(2.8e-13 * units.coulomb),
+    drop_speed_x: Quantity(20 * units.meter / units.second),
+    plate_length_x: Quantity(1.6 * units.centimeter),
+    electric_field_magnitude: Quantity(1.1e6 * units.newton / units.coulomb),
+}
+
+# The vector equation E = F/q can be rewritten as F = Eq. Taking the projection of the electrostatic force
```
blackyblack, line 32: Maybe using vector form of this law would let us avoid writing explanatory comments?

examples/electricity/charged_particle_in_electric_field.py
```diff
@@ -0,0 +1,73 @@
...
+drop_acceleration_y = solve(
+    acceleration_from_force.law,
+    acceleration_from_force.acceleration
+)[0].subs({
+    acceleration_from_force.force: electrostatic_force_y,
+    acceleration_from_force.mass: drop_mass,
+})
+
+# Since no forces act on the drop along the x axis, the projection of the acceleration on it is zero
+horizontal_movement_law = acceleration_law.law.subs({
```
blackyblack, line 52: You can use constant velocity law: https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/kinematic/distance_from_constant_velocity.py

blackyblack (approved):

examples/electricity/charged_particle_in_electric_field.py
alesanter: Using the vector form in fact made these lines much simpler for understanding the code and physics of the process.

examples/electricity/charged_particle_in_electric_field.py
```diff
@@ -0,0 +1,74 @@
...
+drop_mass, drop_charge, drop_speed_x = symbols("drop_mass drop_charge drop_speed_x")
+plate_length_x = symbols("plate_length_x")
+electric_field_magnitude = symbols("electric_field_magnitude")
+
+values = {
+    drop_mass: 1.5e-10,  # kg
+    drop_charge: 2.8e-13,  # C
+    drop_speed_x: 20,  # m/s
+    plate_length_x: 1.6e-2,  # m
+    electric_field_magnitude: 1.1e6,  # N/C
```
blackyblack, line 28: Why did you remove quantites? It was way too more readable before change.

examples/electricity/charged_particle_in_electric_field.py
```diff
@@ -0,0 +1,74 @@
...
+}
+
+electric_field_vector = Vector([0, -electric_field_magnitude, 0])
+electrostatic_force_vector = electric_field_law.electrostatic_force_law(electric_field_vector)
+electrostatic_force_y_expr = electrostatic_force_vector.components[1]
+electrostatic_force_y = (
+    electrostatic_force_y_expr
+    if isinstance(electrostatic_force_y_expr, float)
+    else electrostatic_force_y_expr.subs(electric_field_law.test_charge, -drop_charge)
```
blackyblack, line 37: Why this check? Don't you know what you get in your example?

blackyblack, line 37: Why `-drop_charge`? Simply set negative charge in values.

examples/electricity/charged_particle_in_electric_field.py
```diff
@@ -0,0 +1,74 @@
...
+
+values = {
+    drop_mass: 1.5e-10,  # kg
+    drop_charge: 2.8e-13,  # C
+    drop_speed_x: 20,  # m/s
+    plate_length_x: 1.6e-2,  # m
+    electric_field_magnitude: 1.1e6,  # N/C
+}
+
+electric_field_vector = Vector([0, -electric_field_magnitude, 0])
```
blackyblack, line 31: Pylint does not like `-electric_field_magnitude`. Use `-1 * electric_field_magnitude` instead.

examples/electricity/charged_particle_in_electric_field.py
alesanter: Oh, I mixed up what kind of dimensions I needed to remove.

examples/electricity/charged_particle_in_electric_field.py
alesanter: This check was due to `mypy` failing since `electrostatic_force_vector.components` is of type `list[float | Expr]`. For this reason I cannot call `.subs()` on a vector component since it might be a float.

alesanter: I agree, this is more reasonable.

alesanter: Maybe it would be easier to add the `# type: ignore` comment for `mypy` on this line?

blackyblack: I see. I will find a workaround.

alesanter: For now, I will leave it as is.

blackyblack: I usually use `sympify` to solve it. It converts Python types to sympy expressions. Give it a try.

blackyblack (approved):

PR #234, opened 2024-01-18
Create law and test for work of general force

alesanter:


symplyphysics/laws/dynamics/work_done_by_general_force_in_one_dimension.py
```diff
@@ -0,0 +1,60 @@
...
+    a, b = symbols("a b")
+    solved = solve(
+        [
+            Eq(force_before_, a * position_before_ + b),
+            Eq(force_after_, a * position_after_ + b),
+        ],
+        (a, b),
+        dict=True,
+    )[0]
+    force_function = solved[a] * position + solved[b]
```
blackyblack, line 57: I think there is an easier way to define this function. See here: https://github.com/blackyblack/symplyphysics/blob/25aeded758a35e4643ff2ab53e814c5c202faacd/symplyphysics/definitions/acceleration_is_velocity_derivative.py#L30

symplyphysics/laws/dynamics/work_done_by_general_force_in_one_dimension.py
```diff
@@ -0,0 +1,60 @@
...
+# Law: W = Integral(F(x), x)
+## W - work done by force F
+## F = F(x) - force in question
+## x - position of the object on the axis
+
+work = Symbol("work", units.energy)
+force = Function("force", units.force)
+position = Symbol("position", units.length)
+
+law = Eq(work, Integral(force(position), position))
```
blackyblack, line 27: I think it is better to add integration limits to the law, eg `Integral(force(position), (position, position_start, position_end))`

alesanter: I was unsure at first, but it probably is better to do so indeed.

symplyphysics/laws/dynamics/work_done_by_general_force_in_one_dimension.py
alesanter: But in this form of the function, velocity would be zero when time is zero. The y-intercept of velocity_function does not matter in that case because we take a derivative of it in the law. In this law, however, we integrate the force function, that is why we cannot ignore the intercept. But since I will include both integral limits into the law, these lines will not even be needed.

blackyblack (approved):

PR #247, opened 2024-01-20
Commit #1. potential_of_electrostatic_field

Zulusssss:
Add law and tests.

### DIFF
--- a/symplyphysics/laws/electricity/potential_of_electrostatic_field.py
+++ b/symplyphysics/laws/electricity/potential_of_electrostatic_field.py
@@ -1,0 +1,43 @@
+from sympy import (Eq, solve)
+from symplyphysics import (
+    units,
+    Quantity,
+    Symbol,
+    print_expression,
+    validate_input,
+    validate_output,
+)
+
+# Description
+## Electric potential (also called the electric field potential, potential drop, the electrostatic potential)
+## is defined as the amount of work energy needed per unit of electric charge to move the charge from
+## a reference point to a specific point in an electric field. More precisely, the electric potential is
+## the energy per unit charge for a test charge that is so small that the disturbance of the field under
+## consideration is negligible.
+
+## Law is: ф = W / q, where
+## ф - electrostatic potential,
+## W - potential energy,
+## q - charge.
+
+electrostatic_potential = Symbol("electrostatic_potential", units.voltage)
+
+potential_energy = Symbol("potential_energy", units.energy)
+charge = Symbol("charge", units.charge)
+
+law = Eq(electrostatic_potential, potential_energy / charge)
+
+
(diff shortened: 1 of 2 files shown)


symplyphysics/laws/electricity/potential_of_electrostatic_field.py
```diff
@@ -0,0 +1,43 @@
...
+)
+
+# Description
+## Electric potential (also called the electric field potential, potential drop, the electrostatic potential)
+## is defined as the amount of work energy needed per unit of electric charge to move the charge from
+## a reference point to a specific point in an electric field. More precisely, the electric potential is
+## the energy per unit charge for a test charge that is so small that the disturbance of the field under
+## consideration is negligible.
+
+## Law is: ф = W / q, where
```
blackyblack, line 18: What's the difference with https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/electricity/work_of_charge_transfer_from_charge_and_voltage.py ?

Zulusssss: Here we calculate the potential of the electric field at some point, knowing the potential energy at that point.

https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/electricity/work_of_charge_transfer_from_charge_and_voltage.py
And here we are talking about the amount of energy that must be expended to move the charge between two points with a potential difference of "U".

But both laws are similar in shape and dimensions, yes.

Zulusssss: > Here we calculate the potential of the electric field at some point, knowing the potential energy at that point.
> 
> https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/electricity/work_of_charge_transfer_from_charge_and_voltage.py And here we are talking about the amount of energy that must be expended to move the charge between two points with a potential difference of "U".
> 
> But both laws are similar in shape and dimensions, yes.

Here we calculate the potential of the electric field at some point, knowing the potential energy at that point.

https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/electricity/work_of_charge_transfer_from_charge_and_voltage.py
And here we are talking about the amount of energy that must be expended to move the charge between two points with a potential difference of "U".

But both laws are similar in shape and dimensions, yes.

blackyblack: From your description:
```
is defined as the amount of work energy needed per unit of electric charge to move the charge from a reference point to a specific point in an electric field
```
It is exactly the description of transfer of charge.
I can approve it if you find the description that justifies this law and if you explain potential energy of a field.

blackyblack (approved):

PR #253, opened 2024-01-21
Acceleration of rotating body

alesanter:


symplyphysics/laws/kinematic/vector/acceleration_of_rotating_body.py
```diff
@@ -0,0 +1,45 @@
...
+)
+
+# Description
+## The acceleration of a rotating body is composed of two parts: radial acceleration, which is always
+## present in a rotating environment and which points to the center of rotation, and tangential acceleration,
+## which is responsible for the change of the magnitude of the velocity vector.
+
+# Law: a = a_r + a_t
+## a - total acceleration vector
+## a_r - radial acceleration vector (parallel to position vector of the rotating point)
```
blackyblack, line 17: Probably worth mention that it is opposite to the position vector.

symplyphysics/laws/kinematic/vector/acceleration_of_rotating_body.py
```diff
@@ -0,0 +1,45 @@
...
+## a_t - tangential acceleration vector (perpendicular to position vector of the rotating
+##       point, and tangent to its path)
+
+
+def acceleration_law(
```
blackyblack, line 22: I wonder should we also add functions to calculate tangential vector from sum vector and radial vector? The same for radial from tangential and sum.

blackyblack (approved):

symplyphysics/laws/kinematic/vector/acceleration_of_rotating_body.py
alesanter: It is, will do

symplyphysics/laws/kinematic/vector/acceleration_of_rotating_body.py
alesanter: Hm, since we cannot use sympy's `solve` as we do in scalar laws to find the required components from the equation, this might be necessary. Although I would opt for a less clutter-making alternative if it existed.

blackyblack: I will look to implement some kind of sympy Addition for vectors. Might be quite tricky though.

alesanter: As far as I know, sympy uses `Matrix` for vector manipulation. And in my experience in sympy they mostly only work for the Cartesian coordinates.

test/kinematic/vector/acceleration_of_rotating_body_test.py
```diff
@@ -0,0 +1,136 @@
...
+    ])
+    Args = namedtuple("Args", "a_r a_t a_total")
+    return Args(a_r=a_r, a_t=a_t, a_total=a_total)
+
+
+def test_basic_law(test_args):
+    result = acceleration_law.calculate_acceleration(test_args.a_r, test_args.a_t)
+    assert len(result.components) == 3
+    assert SI.get_dimension_system().equivalent_dims(result.dimension, units.acceleration)
+    correct_values = (-1.0, 0.1, -2.6)
```
blackyblack, line 44: You can take correct values from the `test_args`

blackyblack (approved):

test/kinematic/vector/acceleration_of_rotating_body_test.py
alesanter: It's just that you told me earlier to avoid extra conversions via `convert_to` so I went with this

blackyblack: Yes but you should have a single point of truth. Before you didn't have total vector in test_args so this way you could avoid extra conversions. Now you have and you've got a copypaste - 2 sources of data. It is error prone and should be avoided.

alesanter: I agree with you on this.

blackyblack (approved):

PR #254, opened 2024-01-21
Tangential acceleration of a rotating body

alesanter:


symplyphysics/laws/kinematic/tangential_acceleration_of_rotating_body.py
```diff
@@ -0,0 +1,89 @@
...
+})
+
+# Differentiate both sides of the equation w.r.t. time
+diff_linear_velocity_law = Eq(
+    linear_velocity_law_sub.lhs.diff(time),
+    linear_velocity_law_sub.rhs.diff(time),
+)
+
+# alpha = d(omega)/dt
+diff_linear_velocity_law_sub = diff_linear_velocity_law.subs(
```
blackyblack, line 52: You should use https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/definitions/angular_acceleration_is_angular_velocity_derivative.py to prove that angular acceleration is angular velocity derivative.

symplyphysics/laws/kinematic/tangential_acceleration_of_rotating_body.py
```diff
@@ -0,0 +1,89 @@
...
+    Derivative(angular_velocity(time), time), 
+    angular_acceleration
+)
+
+# a_t = dv/dt
+tangential_acceleration_derived = solve(
```
blackyblack, line 58: You should use https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/definitions/acceleration_is_velocity_derivative.py to prove that linear acceleration is linear velocity derivative.

symplyphysics/laws/kinematic/tangential_acceleration_of_rotating_body.py
```diff
@@ -0,0 +1,89 @@
...
+    diff_linear_velocity_law_sub,
+    Derivative(linear_velocity(time), time)
+)[0]
+
+tangential_acceleration_from_law = solve(law, tangential_acceleration)[0]
```
blackyblack, line 63: You can use `law.rhs` since law structure is known here.

test/kinematic/tangential_acceleration_of_rotating_body_test.py
```diff
@@ -0,0 +1,45 @@
...
+    units,
+    convert_to,
+    Quantity,
+    SI,
+)
+from symplyphysics.laws.kinematic import tangential_acceleration_of_rotating_body as tangential_acceleration_law
+
+# Description
+## A body is moving along a curve. Its angular acceleration is 3.5 rad/s^2 and its rotation radius
+## about the momentary rotation axis is 0.1 m. The magnitude of its tangential acceleration should
```
blackyblack, line 14:
```suggestion
## around the momentary rotation axis is 0.1 m. The magnitude of its tangential acceleration should
```

alesanter: image
_about_ is also possible in this context

blackyblack: Thanks. You could leave it this way as you've proven your point. No need to blindly follow my review :)

alesanter: Oh don't worry, I actually figured the other option was better!

PR #260, opened 2024-01-22
Work done during rotation

alesanter:


symplyphysics/laws/dynamics/mechanical_work_during_rotation.py
```diff
@@ -0,0 +1,88 @@
...
+torque = Symbol("torque", units.force * units.length)
+angular_displacement = Symbol("angular_displacement", angle_type)
+
+law = Eq(work, torque * angular_displacement)
+
+
+# Derive law from non-rotational counterpart.
+## We assume a particle moving along a curved trajectory due to a force F applied to it.
+## The displacement of the particle should be small enough so that the position vector
+## is approximately the same at all points of the trajectory. In that case, the force vector
```
blackyblack, line 36: Probably worth adding to the conditions, that position vector has constant length. Also force should be constant. And angle between force and position vectors should be constant.
Here https://en.wikipedia.org/wiki/Torque there is a proof using integral of force. It seems it does not require force, angle and radius to be constant.

symplyphysics/laws/dynamics/mechanical_work_during_rotation.py
```diff
@@ -0,0 +1,88 @@
...
+## is assumed to form an angle phi with the position vector of the particle. Since only the
+## component of the force vector that is tangent to the path is the one accelerating the
+## particle, it does work on the particle.
+
+force = Symbol("force", units.force)
+angle = Symbol("angle", angle_type)
+radius = Symbol("radius", units.length)
+
+tangent_force = force * sin(angle)
```
blackyblack, line 45: Use projection law. And explain, what is our reference frame. It seems you are looking from the point of the curve perspective, where X axis is radial acceleration and Y is tangential acceleration.

symplyphysics/laws/dynamics/mechanical_work_during_rotation.py
alesanter: I'll take a look at it

symplyphysics/laws/dynamics/mechanical_work_during_rotation.py
```diff
@@ -0,0 +1,98 @@
...
+    torque_def.distance_to_axis: radius,
+    torque_def.angle: angle,
+})
+work_derived_sub = solve(
+    [Eq(work, work_derived), torque_def_sub],
+    (radius, work),
+    dict=True
+)[0][work]
+
+assert expr_equals(work_derived_sub, law.rhs)
```
blackyblack, line 84: I think we can also prove, that it is valid in general case, without conditions that force, angle and radius should be constant.
Here we've proven that on small displacement it works. Now we can say, that small displacement is an infinitesimal amount of work: dW. To get total work, we should integrate: W = integrate(dW, dAngle). dW = dTau * dAngle, but tau is constant (from law definition), so W = tau * integrate(dAngle, angle_start, angle_end) -> W = tau * (angle_end - angle_start)

blackyblack (approved):

symplyphysics/laws/dynamics/mechanical_work_during_rotation.py
alesanter: I can see your point and agree that the radius vector doesn't need to be constant. Although I still think that at least the force (and the angle between it and the radius-vector) should be constant, which are the same constraints as in its linear counterpart $`W = \vec{F} \cdot \vec{s} = F s \cos{\phi}`$, otherwise it would be an integral of $`\vec{F}`$ w.r.t. $`\vec{s}`$.

blackyblack: But more general definition of work is an integral of force.

blackyblack: Therefore with this proof you are adding additional constraints, that should not be here. For example, tau can be constant, while force and radius are not.

alesanter: > Therefore with this proof you are adding additional constraints, that should not be here. For example, tau can be constant, while force and radius are not.

Hm, I agree with you now

symplyphysics/laws/dynamics/mechanical_work_during_rotation.py
```diff
@@ -88,11 +88,14 @@
 
 angle_start = Symbol("angle_start", angle_type)
 angle_end = Symbol("angle_end", angle_type)
+# Infinitesimal work = dTau(angle) * dAngle
```
alesanter, line 91: But since $`W = \tau \theta`$, then $`dW = d(\tau \theta) = \theta d\tau + \tau d\theta`$, not $`d\tau d\theta`$

alesanter: Or you mean that we found infinitesimal work earlier, but in that case we just have $`dW = \tau d\theta`$

blackyblack: Yes, dW = Tau * dTheta

blackyblack: You are right, in description `dTau` should be just `Tau`

symplyphysics/laws/dynamics/mechanical_work_during_rotation.py
```diff
@@ -0,0 +1,124 @@
...
+)[0][work]
+
+assert expr_equals(work_derived_sub, law.rhs)
+
+# Having derived for the infinitesimal case, let us prove this equality in case of a finite angular
+# displacement, with the torque being constant.
+
+angle_start = Symbol("angle_start", angle_type)
+angle_end = Symbol("angle_end", angle_type)
+# Infinitesimal work = dTau(angle) * dAngle
```
blackyblack, line 91:
```suggestion
# Infinitesimal work = Tau(angle) * dAngle
```

PR #261, opened 2024-01-22
Create example for average angular velocity

alesanter:


examples/kinematics/average_angular_velocity_during_dive.py
```diff
@@ -0,0 +1,39 @@
...
+    const_acceleration_law.distance(const_acceleration_law.movement_time): height,
+})
+# First solution is negative
+dive_time = solve(const_acceleration_law_sub, const_acceleration_law.movement_time)[1]
+
+angular_velocity_law_sub = angular_velocity_law.law.subs({
+    angular_velocity_law.radians: total_angular_displacement,
+    angular_velocity_law.time: dive_time,
+})
+average_angular_velocity = solve(
```
blackyblack, line 27: Add explanation, why frequency is an average velocity.

blackyblack (approved):

examples/kinematics/average_angular_velocity_during_dive.py
alesanter: Are angular frequency and angular frequency not synonyms? Although they do sound as if they are not

blackyblack: It seems they are. Although frequency is absolute value of velocity. Also taking velocity for average velocity only works when it is constant.

alesanter: First, in this example we are not interested in the sign of the velocity, only in its absolute value, so to me it is quite reasonable to for them to be interchangeable here. Next, the usual formula for finding average velocity is average displacement over change in time ($`\overline{v} = \frac {\Delta x} {\Delta t}`$), and here I take the same approach: $`\overline{\omega} = \frac {\Delta \theta} {\Delta t}`$.

blackyblack: Yes but I want to see it in comments in the code, not in github review.

alesanter: Oh, I see what you mean

blackyblack (approved):

PR #277, opened 2024-01-24
Commit #1. ampere_law

Zulusssss:
Add law and tests.

### DIFF
--- a/symplyphysics/laws/electricity/ampere_law.py
+++ b/symplyphysics/laws/electricity/ampere_law.py
@@ -6,7 +6,7 @@ from symplyphysics import (
     print_expression,
     validate_input,
     validate_output,
-    dimensionless
+    angle_type
 )
 
 # Description
@@ -21,12 +21,16 @@ from symplyphysics import (
 ## B - magnetic induction,
 ## a - angle between magnetic induction and current direction.
 
+# Conditions:
+## - Сonductor must be thin;
+## - Сonductor must be of finite length.
+
 force = Symbol("force", units.force)
 
 current = Symbol("current", units.current)
 length = Symbol("length", units.length)
 induction = Symbol("induction", units.magnetic_density)
-angle = Symbol("angle", dimensionless)
+angle = Symbol("angle", angle_type)
 
 law = Eq(force, current * length * induction * cos(angle))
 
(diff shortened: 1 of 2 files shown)


symplyphysics/laws/electricity/ampere_law.py
```diff
@@ -0,0 +1,51 @@
...
+## L - length of conductor,
+## B - magnetic induction,
+## a - angle between magnetic induction and current direction.
+
+force = Symbol("force", units.force)
+
+current = Symbol("current", units.current)
+length = Symbol("length", units.length)
+induction = Symbol("induction", units.magnetic_density)
+angle = Symbol("angle", dimensionless)
```
blackyblack, line 29: You can use `angle_type` here, I think.

symplyphysics/laws/electricity/ampere_law.py
```diff
@@ -0,0 +1,51 @@
...
+    validate_output,
+    dimensionless
+)
+
+# Description
+## Ampere's law is the law that determines the force with which a magnetic field acts on
+## a small segment of a conductor with a current. The force turns out to be linearly dependent 
+## on both current and magnetic induction.
+
+## Law is: F = I * L * B * cos(a), where
```
blackyblack, line 17: Are there any requirements to the conductor geometry. Put them in conditions, if there are any.

symplyphysics/laws/electricity/ampere_law.py
```diff
@@ -0,0 +1,51 @@
...
+def print_law() -> str:
+    return print_expression(law)
+
+
+@validate_input(current_=current,
+    length_=length,
+    angle_=angle,
+    induction_=induction)
+@validate_output(force)
+def calculate_force(current_: Quantity, length_: Quantity, angle_: float, induction_: Quantity) -> Quantity:
```
blackyblack, line 43: You can accept angle as both Quantity and float, because Quantity allows some angle conversions.

blackyblack (approved):

symplyphysics/laws/electricity/ampere_law.py
Zulusssss: > You can use `angle_type` here, I think.

I can't find "angle_type" in the "units" module.

blackyblack: It's in `symplyphysics`:
`from symplyphysics import angle_type`

blackyblack (approved):

PR #289, opened 2024-01-25
Law and test for LJ potential

alesanter:


symplyphysics/laws/chemistry/lennard_jones_potential.py
```diff
@@ -0,0 +1,53 @@
+from sympy import Eq
```
blackyblack, line 1: Don't use marketing names for laws.

symplyphysics/laws/chemistry/lennard_jones_potential.py
```diff
@@ -0,0 +1,53 @@
...
+## other at a very close distance, attract each other at moderate distance, and do not interact at infinite
+## distance.
+
+# Law: U = 4*e*((sigma/r)**12 - (sigma/r)**6)
+## U - Lennard-Jones potential
+## e - depth of potential well ("dispersion energy")
+## sigma - distance at which the potential energy is zero ("particle size")
+## r - distance between interacting particles
+
+lj_potential = Symbol("potential", units.energy)
```
blackyblack, line 24: Don't shortcut and use descriptive name.

test/chemistry/lennard_jones_potential_test.py
```diff
@@ -0,0 +1,50 @@
...
+
+# Description
+## The value of the Lennard-Jones potential for two interacting Xenon atoms at the distance of
+## 1 Å is 2.65e-13 J (e = 2.94e-21 J, sigma = 4.1 Å)
+
+
+@fixture(name="test_args")
+def test_args_fixture():
+    e = Quantity(2.94e-21 * units.joule)
+    sigma = Quantity(4.10e-10 * units.angstrom)
```
blackyblack, line 20: Number does not match with description.

symplyphysics/laws/chemistry/lennard_jones_potential.py
alesanter: There are lots of other potentials that target the same topic (Mie, Buckingham, Morse potentials) though

alesanter: Maybe `lennard_jones_intermolecular_pair_potential`?

blackyblack: Hm, good point. I suggest to add folder, eg `potential_energy_models` and put different models in it. This way we can use fancy model names without cluttering the project.

test/chemistry/lennard_jones_potential_test.py
```diff
@@ -17,7 +17,7 @@
 @fixture(name="test_args")
 def test_args_fixture():
     e = Quantity(2.94e-21 * units.joule)
-    sigma = Quantity(4.10e-10 * units.angstrom)
+    sigma = Quantity(4.10 * units.angstrom)
```
blackyblack, line 20: How is it possible, that 10 orders of magnitude didn't change the test results?

symplyphysics/laws/chemistry/lennard_jones_potential.py
alesanter: I agree

test/chemistry/lennard_jones_potential_test.py
alesanter: I'm surprised myself. I'll look into it

symplyphysics/laws/chemistry/lennard_jones_potential.py
blackyblack: I think folder should have `__init__.py` to work properly. Tests don't need it.

symplyphysics/laws/chemistry/potential_energy_models/lennard_jones_potential.py
```diff
@@ -0,0 +1,53 @@
...
+## sigma - distance at which the potential energy is zero ("particle size")
+## r - distance between interacting particles
+
+potential = Symbol("potential", units.energy)
+dispersion_energy = Symbol("dispersion_energy", units.energy)
+particle_size = Symbol("particle_size", units.length)
+distance = Symbol("distance", units.length)
+reduced_distance = distance / particle_size
+
+law = Eq(potential, 4 * dispersion_energy * (reduced_distance**(-12) - reduced_distance**(-6)))
```
blackyblack, line 30: Why did you invert the formula? Now it does not look like the law in description.

test/chemistry/lennard_jones_potential_test.py
alesanter: Perhaps, this is the behaviour of `approx`. With the correct `sigma` value of `4.10 Å`, the potential value is `2.65e-13 J` as expected, and with the wrong value of `4.10e-10 Å`, the potential value calculated is `-5.59e-77 J`. The documentation for `approx` says that it uses either `math.isclose` or `numpy.isclose` under the hood. Although only `numpy` gives `True`
image

blackyblack: Use it like this:
`assert result_value == approx(2.65e-13, abs=1e-15)`
For some reason, tolerance is overriden by some default value, unless it is absolute.

alesanter: I also came to that conclusion

symplyphysics/laws/chemistry/potential_energy_models/lennard_jones_potential.py
blackyblack: Simply remove reduced distance and write law in the same form as in description.

blackyblack (approved):

PR #292, opened 2024-01-25
Commit #1. radius_of_a_charged_particle_in_a_magnetic_field

Zulusssss:
Add law and tests.


symplyphysics/laws/electricity/radius_of_a_charged_particle_in_a_magnetic_field.py
```diff
@@ -0,0 +1,46 @@
...
+    print_expression,
+    validate_input,
+    validate_output,
+)
+
+# Description
+## Let an arbitrary particle move in a magnetic field around a circle. Then radius of its motion depends on
+## magnetic induction and on velocity, charge and mass of the particle.
+
+## Law is: r = m * v / (q * B), where
```
blackyblack, line 15: Do we really need this law. You simply substituted `T` for `2pi * r / v` in https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/electricity/period_of_a_charged_particle_in_a_magnetic_field.py

Zulusssss: > Do we really need this law. You simply substituted `T` for `2pi * r / v` in

I don't quite understand the remark.
Yes, these laws are related, but they give different values and have a similar, but not identical appearance.

I will also give examples of laws from the repository in which the similarity with each other is either the same or higher:
1) https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/waves/wavespeed_from_medium.py

https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/waves/wavespeed_from_medium_permittivity_permeability.py

2) https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/waves/photon_momentum_is_proportional_to_frequency.py

https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/waves/photon_momentum_is_proportional_to_propagation_vec.py

blackyblack: Yes, sometimes we add very closely related laws. But for a reason: they are widely used in this form; or it was a practice task. If you think this form is better for some applications, or better known for some users, it's ok to keep it.
Anyway, I expect you to derive one form from another, if you decide to keep it.

Zulusssss: > Yes, sometimes we add very closely related laws. But for a reason: they are widely used in this form; or it was a practice task. If you think this form is better for some applications, or better known for some users, it's ok to keep it. Anyway, I expect you to derive one form from another, if you decide to keep it.

Ok, got it. I'll add a proof.

symplyphysics/laws/electricity/radius_of_a_charged_particle_in_a_magnetic_field.py
```diff
@@ -28,6 +30,23 @@
 
 law = Eq(radius, mass * velocity / (charge * induction))
 
+# This law might be derived via period of a charged particle in a magnetic field.
+
+#TODO:
+## The following law should be added:
+## T = 2 * pi * radius / velocity
```
blackyblack, line 37: Well, we have `T = 2 * pi / w` law, so only `w = v / r` missing. But you can simply use `distance_from_constant_velocity` law, given distance is a full circle = `2 * pi * r` and velocity is constant. 
But you should mention the definition of period: it is a time taken to cover a full circle.

symplyphysics/laws/electricity/radius_of_a_charged_particle_in_a_magnetic_field.py
```diff
@@ -1,4 +1,4 @@
-from sympy import (Eq, solve, pi)
+from sympy import (Eq, solve, pi, solveset, dsolve)
```
blackyblack, line 1: solveset and dsolve are unused

blackyblack (approved):

blackyblack (approved):

PR #294, opened 2024-01-25
volume_charge_density_from_charge_and_volume

Darya2909:


symplyphysics/definitions/volume_charge_density_from_charge_and_volume.py
```diff
@@ -0,0 +1,37 @@
+from sympy import (Eq, solve)
+from symplyphysics import (units, Quantity, Symbol, print_expression, validate_input,
+                           validate_output)
+
+# Description
+## The volume charge density is defined as the amount of charge present over a unit volume of the conductor.
+## Charge density can be either positive or negative, since electric charge can be either positive or negative.
+
+
+## Definition: ρ = q / V
```
blackyblack, line 10: Derive this law from https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/definitions/volume_number_density.py

Darya2909: This is a definition, not a law, it is not derived from anywhere

blackyblack: Yes, but obviously they are the same formula. So we can say, that by definition, volumetric charge density is volumetric density (which we have a definition of N/V) of charges. And this specific formula in your law is derived from the definitions of volumetric density and of volumetric charge density.

symplyphysics/definitions/volume_charge_density_from_charge_and_volume.py
```diff
@@ -22,6 +24,25 @@
...
 
+## Proof
+
+## Let's express the number of objects from the volume number density definition
+## In this case, the objects are charged particles.
+expr = volume_number_density.definition.subs({
+        volume_number_density.volume: volume,
+    })
+
+number_of_particles = solve(expr,  volume_number_density.objects, dict=True)[0][volume_number_density.objects]
```
blackyblack, line 35: Simply substitute `objects` for `charge`

symplyphysics/definitions/volume_charge_density_from_charge_and_volume.py
```diff
@@ -0,0 +1,50 @@
...
+## ρ is volume charge density
+## q is charge
+## V is the volume over which charge is distributed
+
+
+volume_charge_density = Symbol("volume_charge_density", units.charge / units.volume)
+charge = Symbol("charge", units.charge)
+volume = Symbol("volume", units.volume)
+
+definition = Eq(volume_charge_density, charge / volume)
```
blackyblack, line 22: Since now it has proof, move this law outside of definitions.

blackyblack (approved):

PR #297, opened 2024-01-26
Proof of work-energy principle

alesanter:


symplyphysics/laws/dynamics/total_work_equals_change_in_kinetic_energy.py
```diff
@@ -25,6 +35,64 @@
 law = Eq(total_work, kinetic_energy(time_after) - kinetic_energy(time_before))
 
 
+# Derive the law in case of rectilinear motion with constant total force acting on particle
```
blackyblack, line 38: Probably also should be integrated (over time, I guess) to remove constant force condition.

PR #309, opened 2024-01-28
Commit #1. light_interference_in_thin_films

Zulusssss:
Add law and tests.


symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
```diff
@@ -0,0 +1,57 @@
...
+## Interference in thin films is a phenomenon that occurs as a result of the separation of a light beam when
+## reflected from the upper and lower boundaries of a thin film. As a result, there are two light waves that
+## can interfere.
+## The order of interference indicates the number of whole wavelengths. When the difference in the path of the
+## beam reflected from the upper boundary of the film and the beam reflected from the lower boundary of the
+## film is equal to an integer wave "n", then these two waves enter the observation point with the same phases
+## and demonstrate interference.
+## The angle of refraction of light is the angle between the ray passing through the surface and the normal to the surface.
+
+## Law is: h = k * L / (2 * n * sin(O)), where
```
blackyblack, line 23: Could you show me, where you've got this formula?

Zulusssss: > Could you show me, where you've got this formula?

http://fn.bmstu.ru/data-physics/library/physbook/tom4/ch4/texthtml/ch4_4.htm

I have already noticed that there should be a cosine. I'll fix it.

blackyblack: Do you mean formula 4.24b: 2nd * cos(theta2) = mL?
But it is a condition of minimum of interference. You've never mention that we expect minimum of interference in the law.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
```diff
@@ -0,0 +1,57 @@
...
+
+# Description
+## Interference in thin films is a phenomenon that occurs as a result of the separation of a light beam when
+## reflected from the upper and lower boundaries of a thin film. As a result, there are two light waves that
+## can interfere.
+## The order of interference indicates the number of whole wavelengths. When the difference in the path of the
+## beam reflected from the upper boundary of the film and the beam reflected from the lower boundary of the
+## film is equal to an integer wave "n", then these two waves enter the observation point with the same phases
+## and demonstrate interference.
+## The angle of refraction of light is the angle between the ray passing through the surface and the normal to the surface.
```
blackyblack, line 21: Angle of refraction is angle between refracted ray and the normal.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
```diff
@@ -0,0 +1,57 @@
...
+
+## Law is: h = k * L / (2 * n * cos(O)), where
+## h - film thickness,
```
blackyblack, line 24: What is this thickness that we've got here? What does it mean?

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
```diff
@@ -0,0 +1,57 @@
...
+## k - order of interference,
+## L - wavelength,
```
blackyblack, line 26: Wavelength in medium or in vacuum?

test/optics/interference_films_from_angle_refraction_refractive_index_wavelength_test.py
```diff
@@ -0,0 +1,60 @@
...
+    with raises(TypeError):
+        thickness_law.calculate_film_thickness(test_args.wavelength, test_args.order_interference, True, test_args.angle_refraction)
+
+
+def test_bad_angle_refraction(test_args):
+    angle_refraction = Quantity(1 * units.coulomb)
+    with raises(errors.UnitsError):
+        thickness_law.calculate_film_thickness(test_args.wavelength, test_args.order_interference, test_args.refractive_index, angle_refraction)
+    with raises(AttributeError):
+        thickness_law.calculate_film_thickness(test_args.wavelength, test_args.order_interference, test_args.refractive_index, True)
```
blackyblack, line 60: Don't use True for args. Sympy complains about deprecated method when you do it.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
```diff
@@ -0,0 +1,57 @@
...
+    validate_output,
+    dimensionless,
+    angle_type
+)
+
+# Description
+## Interference in thin films is a phenomenon that occurs as a result of the separation of a light beam when
+## reflected from the upper and lower boundaries of a thin film. As a result, there are two light waves that
+## can interfere.
+## The order of interference indicates the number of whole wavelengths. When the difference in the path of the
```
blackyblack, line 17: You can mention that order of interference can be chosen arbitrarily, to achieve the desired thin film thickness.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
Zulusssss: > Do you mean formula 4.24b: 2nd * cos(theta2) = mL? But it is a condition of minimum of interference. You've never mention that we expect minimum of interference in the law.

If instead of m we substitute m = 1, 2, 3, ..., then this is the minimum condition. If instead of m we substitute m = 1/2, 3/2, 5/2, ..., then this is the maximum condition.

I can add this to the description. Is it okay?

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
blackyblack: Yes, but what does this thickness mean? What happens if it is 1 centimeter?

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
blackyblack: Not really. You've set the order of interference type to `int` so we cannot substitute it for non integer number. Therefore you describe the minimum condition in your law.
We can set order of interference to `float` though. This way we can describe both minumum and maximum conditions with this law. But what if someone puts some number that is not 1/2 or 1, ie 0.3?
We have two options here:
1. Throw an exception if order of interference is not integer and is not divisible by 2.
2. Use half wavelengths and always use integer number for order of interference. But it may confuse users who expected full wavelengths in this formula.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
Zulusssss: > Yes, but what does this thickness mean? What happens if it is 1 centimeter?

Depends on other values. The interference pattern may not change at all, it may become the exact opposite (maxima will change with minima), or it may change arbitrarily.

Zulusssss: > Yes, but what does this thickness mean? What happens if it is 1 centimeter?

At the moment, I completely do not understand the essence of the remark and what needs to be corrected.

blackyblack: The reason for this remark is that law should have any usability for the users. What's the point having a law, that gives back some arbitrary thickness? You could as well just generate some random output.
You have some calculations that give to user some thickness. That's why I am asking, what is this thickness? Why is it important?

Zulusssss: > The reason for this remark is that law should have any usability for the users. What's the point having a law, that gives back some arbitrary thickness? You could as well just generate some random output.
> You have some calculations that give to user some thickness. That's why I am asking, what is this thickness? Why is it important?

With the interference order equal to 1, with other parameters set, this thickness will be minimal in order to see some kind of interference pattern. That is, for a fixed material, the angle of refraction and wavelength, it is possible to determine the minimum thickness of the film so that an interference pattern is observed.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
```diff
@@ -46,7 +47,13 @@ def print_law() -> str:
     refractive_index_=refractive_index,
     angle_refraction_=angle_refraction)
 @validate_output(film_thickness)
-def calculate_film_thickness(wavelength_: Quantity, order_interference_: int, refractive_index_: float, angle_refraction_: float | Quantity) -> Quantity:
+def calculate_film_thickness(wavelength_: Quantity, order_interference_: int | float, refractive_index_: float, angle_refraction_: float | Quantity) -> Quantity:
+    check = order_interference_ * 2 % 2
```
blackyblack, line 51: On a second thought, it looks like a bad idea. Float division can give quite unpredictable results, so we may face weird issues when trying to distinguish 0.5 from 0.499995.
Let's accept integer number here, but rename it to `double_order_interference` or `number_of_half_waves`. Your newly added checks can be removed, but tests should be updated accordingly to new parameter meaning.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
blackyblack: Are you sure? I see something about minimum of interference is visible, when number of half waves is even. Shouldn't some kind of interference pattern be visible with any order of interference number?

Zulusssss: > Are you sure? I see something about minimum of interference is visible, when number of half waves is even. Shouldn't some kind of interference pattern be visible with any order of interference number?

There is a condition for the minimum interference pattern at the observation point at infinity (4.24b):
http://fn.bmstu.ru/data-physics/library/physbook/tom4/ch4/texthtml/ch4_4.htm

We will set all the parameters except the thickness of the film. Then, for some film values that are less than the minimum thickness value, the equality of 4.24b simply will not be fulfilled. If the interference order was set to 1, then for large parameters this equality will not be fulfilled for thickness values less than the minimum. If the interference order is arbitrary and equal to "m", then condition 4.24 will not be satisfied for all values of the order from m to infinity with a width less than the minimum.

That's what I meant.

blackyblack: It is not a minimal thickness to see interference pattern. It is thickness that gives you interference minimum at a certain point on the interference pattern. So you should update law name and description accordingly.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
```diff
@@ -0,0 +1,61 @@
...
+)
+
+# Description
+## Interference in thin films is a phenomenon that occurs as a result of the separation of a light beam when
+## reflected from the upper and lower boundaries of a thin film. As a result, there are two light waves that
+## can interfere.
+## The order of interference indicates the number of whole wavelengths. When the difference in the path of the
+## beam reflected from the upper boundary of the film and the beam reflected from the lower boundary of the
+## film is equal to an integer wave "n", then these two waves enter the observation point with the same phases
+## and demonstrate interference.
```
blackyblack, line 20: It is wrong. Same phase is not necessary condition to observe interference.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
```diff
@@ -0,0 +1,61 @@
...
+## h - film thickness,
+## k - order of interference,
+## L - wavelength in vacuum,
+## n - refractive index of film
+## O - angle of refraction.
+
+film_thickness = Symbol("film_thickness", units.length)
+
+wavelength = Symbol("wavelength", units.length)
+double_order_interference = Symbol("double_order_interference", dimensionless)
```
blackyblack, line 34: There is nothing about double order of interference in the description. You should update description accordingly.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
```diff
@@ -0,0 +1,61 @@
...
+    return print_expression(law)
+
+
+@validate_input(wavelength_=wavelength,
+    double_order_interference_=double_order_interference,
+    refractive_index_=refractive_index,
+    angle_refraction_=angle_refraction)
+@validate_output(film_thickness)
+def calculate_film_thickness(wavelength_: Quantity, double_order_interference_: int, refractive_index_: float, angle_refraction_: float | Quantity) -> Quantity:
+    if double_order_interference_ <= 0:
```
blackyblack, line 51: Why `0` is forbidden?

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
```diff
@@ -0,0 +1,61 @@
...
+        raise ValueError("double_order_interference_ must be greater than 0.")
```
blackyblack, line 52:
```suggestion
        raise ValueError("Double_order_interference_ must be greater than 0.") 
```

test/optics/interference_films_from_angle_refraction_refractive_index_wavelength_test.py
```diff
@@ -0,0 +1,56 @@
+from collections import namedtuple
+from pytest import approx, fixture, raises
+from sympy import pi
+from symplyphysics import (units, SI, convert_to, Quantity, errors)
+from symplyphysics.laws.optics import interference_films_from_angle_refraction_refractive_index_wavelength as thickness_law
+
+# Description
+## Let the wavelength be 1500 nanometer, the order of interference is 1, the angle of refraction is
+## 60 degree (pi / 3 radian), and the refractive index is 5. Then the film thickness will be 300 nanometer.
+## https://www.indigomath.ru//raschety/uHv0Ia.html
```
blackyblack, line 10: Use some external example. This synthetic example does not verify your understanding of the law.

test/optics/interference_films_from_angle_refraction_refractive_index_wavelength_test.py
```diff
@@ -0,0 +1,56 @@
...
+    wavelength = Quantity(1500 * units.nanometer)
+    double_order_interference = 1
+    angle_refraction = pi / 3
+    refractive_index = 5
+
+    Args = namedtuple("Args", ["wavelength", "double_order_interference", "angle_refraction", "refractive_index"])
+    return Args(wavelength=wavelength, double_order_interference=double_order_interference, refractive_index=refractive_index, angle_refraction=angle_refraction)
+
+
+def test_basic_thickness(test_args):
```
blackyblack, line 24: Add types.

test/optics/interference_films_from_angle_refraction_refractive_index_wavelength_test.py
```diff
@@ -0,0 +1,56 @@
...
+    result = thickness_law.calculate_film_thickness(test_args.wavelength, test_args.double_order_interference, test_args.refractive_index, test_args.angle_refraction)
+    assert SI.get_dimension_system().equivalent_dims(result.dimension, units.length)
+    result = convert_to(result, units.nanometer).evalf(5)
+    assert result == approx(300, rel=0.1)
```
blackyblack, line 28: Use `assert_equal`

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
Zulusssss: > Why `0` is forbidden?

If we put k = 0 in the formula, then the thickness will be equal to 0. It doesn't make physical sense.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
Zulusssss: > It is wrong. Same phase is not necessary condition to observe interference.

Single-phase and antiphase waves are literally what makes the interference pattern, because single-phase waves add up to each other, forming intensity maxima, and antiphase waves add up to each other, forming minima.

blackyblack: Yes, but pattern itself has many phase shifts. When you look at certain point, it appears after some phase shift. It may be minimum, maximum or something in the middle. So your law is about minimum condition that is about certain point of interference pattern.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
Zulusssss: > It is not a minimal thickness to see interference pattern. It is thickness that gives you interference minimum at a certain point on the interference pattern. So you should update law name and description accordingly.

In many tasks that I have seen, they ask me to find the minimum film thickness at which a maximum or minimum intensity will be observed for a given wavelength and angle of refraction. I do not understand any other explanation of the meaning of thickness, other than the one I described above. Therefore, I do not understand how I should change the name of the law and its description.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
Zulusssss: > So your law is about minimum condition that is about certain point of interference pattern.

Yes. And at this point, the waves reflected from the first and second surfaces of the film add up in antiphase, creating an interference minimum.

blackyblack: So you law is about interference minimum. Which is not reflected neither in law name nor in the description.

symplyphysics/laws/optics/interference_films_from_angle_refraction_refractive_index_wavelength.py
blackyblack: You are messing two things: interference minimum and minumal thickness. With your law you can find thickness that results in interference minimum at a certain point. It does not necessarily results in minimal thickness. For example by changing the order of interference you can change the resulting thickness so it is no longer minimal.

test/optics/interference_films_from_angle_refraction_refractive_index_wavelength_test.py
blackyblack: Well you've found somewhat valid example but numbers do not match.

test/optics/film_thickness_for_minimum_interference_from_angle_refraction_refractive_index_wavelength_test.py
```diff
@@ -0,0 +1,54 @@
...
+    order_interference = 4
+    angle_refraction = 0.12 * pi
+    refractive_index = 1.33
+
+    return Args(wavelength=wavelength, order_interference=order_interference, refractive_index=refractive_index, angle_refraction=angle_refraction)
+
+
+def test_basic_thickness(test_args: Args) -> None:
+    result = thickness_law.calculate_film_thickness(test_args.wavelength, test_args.order_interference, test_args.refractive_index, test_args.angle_refraction)
+    assert_equal(result, 647 * units.nanometer)
```
blackyblack, line 26: Does not correspond to description.

blackyblack (approved):

PR #317, opened 2024-01-29
Derive linear displacement during rotation

alesanter:


symplyphysics/laws/kinematic/angular_position_is_arc_length_over_radius.py
```diff
@@ -27,6 +30,38 @@
...
+    path_radius: radius_norm,
+})
+
+# From [the law](../linear_displacement_is_angular_displacement_cross_radius.py), taking the vector
+# norm of both sides, we have norm(s) = norm(theta) * norm(r) * sin(phi) where s is the linear
+# displacement vector, theta is the angular displacement pseudovector, r is the radius vector, and
+# phi is the and between the last two. As a necessary condition, theta and r are orthogonal to each
+# other, therefore phi is pi/2 and sin(phi) is 1, and so abs(s) = abs(theta) * abs(r)
+
+displacement_derived = angle_norm * radius_norm
```
blackyblack, line 60: You should derive it from the law, not hardcode it.

blackyblack (changes requested):

symplyphysics/laws/kinematic/angular_position_is_arc_length_over_radius.py
alesanter: Well, that was a sly hack of mine, I agree. I will try and add laws for cross and dot products since they are often needed but there is nothing to refer to in code.

blackyblack: But you're referring cross product law. Why not use it for proof?

alesanter:
Opened #376

alesanter:
Closed in favor of #1095.

PR #334, opened 2024-01-31
Commit #1. Maximum height of a body thrown at an angle to horizon

Zulusssss:
Add law and tests.

### DIFF
--- a/symplyphysics/laws/gravity/maximum_height_of_a_body_thrown_at_an_angle_to_horizon.py
+++ b/symplyphysics/laws/gravity/maximum_height_of_a_body_thrown_at_an_angle_to_horizon.py
@@ -1,0 +1,64 @@
+from sympy import (Eq, solve, sin, pi)
+from sympy.physics.units import acceleration_due_to_gravity as earth_free_fall_acceleration
+from symplyphysics import (units, Quantity, Symbol, print_expression, validate_input,
+    validate_output, angle_type)
+from symplyphysics.core.expr_comparisons import expr_equals
+from symplyphysics.laws.kinematic import constant_acceleration_movement_is_parabolic as distance_law
+from symplyphysics.laws.kinematic import planar_projection_is_cosine as projection_law
+
+# Description
+## Let's say we throw the body at an angle to the horizon with some initial velocity.
+## Then the time of movement (time of fall) of the body will depend only on the initial velocity,
+## the angle of the throw and the acceleration of free fall.
+## The angle of the throw is the angle between the initial velocity vector and the horizontal axis.
+
+## Law is: t = 2 * v0 * sin(a) / g, where
+## t - movement time,
+## v0 - initial velocity,
+## a - angle of throw,
+## g - acceleration of free fall.
+
+height = Symbol("height", units.length)
+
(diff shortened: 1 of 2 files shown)


symplyphysics/laws/gravity/maximum_height_of_a_body_thrown_at_an_angle_to_horizon.py
```diff
@@ -0,0 +1,70 @@
...
+    projection_law.vector_length: initial_velocity,
+    projection_law.vector_angle: (pi / 2) - angle,
+})
+projection_derived = solve(projection_law_applied, projection_law.projection, dict=True)[0][projection_law.projection]
+
+time_law_applied = time_law.law.subs({
+    time_law.initial_velocity : initial_velocity,
+    time_law.angle: angle,
+})
+time_derived = solve(time_law_applied, time_law.movement_time, dict=True)[0][time_law.movement_time]
```
blackyblack, line 44: You've just made a law to calculate this. Maybe reuse it?

symplyphysics/laws/gravity/maximum_height_of_a_body_thrown_at_an_angle_to_horizon.py
```diff
@@ -0,0 +1,70 @@
...
+
+# The acceleration of gravity is directed opposite to the vertical coordinate axis,
+## so there is a minus sign before the acceleration.
+height_law_applied = distance_law.law.subs({
+    distance_law.initial_velocity: projection_derived,
+    distance_law.movement_time: time_derived / 2,
```
blackyblack, line 50: You cannot just divide it to two. You can use the fact, that vertical velocity is zero in the highest point of trajectory. Or that movement function derivative is zero at the maximum point.

Zulusssss: > You cannot just divide it to two. You can use the fact, that vertical velocity is zero in the highest point of trajectory. Or that movement function derivative is zero at the maximum point.

I can divide the total flight time into two, since the flight path is parabolic and the parabola is symmetrical. Can I add this to the comment?

symplyphysics/laws/gravity/maximum_height_of_a_body_thrown_at_an_angle_to_horizon.py
Zulusssss: > You've just made a law to calculate this. Maybe reuse it?

So I'm overusing the law "maximum_movement_time_of_a_body_thrown_at_an_angle_to_horizon" here.
Is it necessary to reuse it in some other way?

blackyblack: Sorry, I've missed it

symplyphysics/laws/gravity/maximum_height_of_a_body_thrown_at_an_angle_to_horizon.py
blackyblack: We should avoid assumptions if possible. You can derive this time without any assumptions.

PR #336, opened 2024-01-31
total_voltage_with_series_connection_of_capacitors

Darya2909:


symplyphysics/laws/electricity/total_voltage_with_series_connection_of_capacitors.py
```diff
@@ -0,0 +1,36 @@
...
+    validate_output)
+from symplyphysics.core.operations.sum_array import SumArray
+from symplyphysics.core.symbols.symbols import tuple_of_symbols
+from typing import Sequence
+
+# Description
+## When the capacitors are connected in series, the voltage is equal to the sum of the voltages on the individual capacitors.
+## A series connection is used to increase the breakdown voltage of the capacitors.
+
+## Law: V = sum(V_i)
```
blackyblack, line 12: This is exactly the Kirchhoff law for voltages: https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/electricity/circuits/sum_of_all_voltages_in_loop_is_zero.py
What if instead you add a law of voltage of a single capacitor from a series?

blackyblack (changes requested):

PR #338, opened 2024-02-01
Law: rotational inertia of slab

alesanter:


symplyphysics/laws/kinematic/rotational_inertia/geometries/slab_about_perpendicular_axis_through_center.py
```diff
@@ -0,0 +1,72 @@
...
+
+height = Symbol("height", units.length)
+volume = length * width * height
+
+density = density_def.definition.rhs.subs({
+    density_def.mass: mass,
+    density_def.volume: volume,
+})
+
+distance_to_axis = sqrt(integral_law.x**2 + integral_law.y**2)
```
blackyblack, line 43: Mention that you rotate around Z-axis, which is also the axis, where height is measured.

blackyblack (approved):

PR #339, opened 2024-02-01
Law and proof: rotational inertia of thin rod

alesanter:


symplyphysics/laws/kinematic/rotational_inertia/geometries/thin_rod_about_axis_through_center_perpendicular_to_length.py
```diff
@@ -0,0 +1,77 @@
...
+## L - length of rod
+
+rotational_inertia = Symbol("rotational_inertia", units.mass * units.length**2)
+mass = Symbol("mass", units.mass)
+length = Symbol("length", units.length)
+
+law = Eq(rotational_inertia, mass * length**2 / 12)
+
+
+# Derive law from definition of rotational inertia
```
blackyblack, line 30: Why not take the inertia of a slab and assume its width is zero? I think it is equivalent to a thin rod, and even to a more general case of a rotating plane.

symplyphysics/laws/kinematic/rotational_inertia/geometries/thin_rod_about_axis_through_center_perpendicular_to_length.py
```diff
@@ -0,0 +1,77 @@
...
+# its parts, namely, by integrating over the rod's length.
+
+infinitesimal_length = SymSymbol("infinitesimal_length", positive=True)
+infinitesimal_mass = SymSymbol("infinitesimal_mass", positive=True)
+
+# Because the rod is uniform, the ratio of mass to length is equal for all elements of
+# the rod and for the rod itself. This equation basically states that the local linear
+# density of the rod is constant and equal to the ratio of rod's mass to its length.
+
+uniformity_eqn = Eq(infinitesimal_mass / infinitesimal_length, mass / length)
```
blackyblack, line 43: You could derive it from the constant density condition. But for constant density you could use `rotational_inertia_cartesian_integral` I guess, without integrating manually.

symplyphysics/laws/kinematic/rotational_inertia/geometries/thin_rod_about_axis_through_center_perpendicular_to_length.py
alesanter: My goal was mostly to show how to use the most basic definition to find the integral rotational inertia, which might be useful more more complicated cases. But I agree, it's a subcase of a slab with zero width.

blackyblack: It's hard to expect someone would  find this law, when trying to find out how to integrate over rotating body. Let's use easy way to prove it and use complex way, when it's unavoidable, eg some weird body form.

blackyblack:
You have all recent commits in your PR. Merge or rebase to fix.

blackyblack (approved):

PR #358, opened 2024-02-06
Core: equivalent_dims

alesanter:
This core function is supposed to simplify the boilerplate code in tests and might be helpful in general use too.

### DIFF
--- a/test/chemistry/potential_energy_models/lennard_jones_potential_test.py
+++ b/test/chemistry/potential_energy_models/lennard_jones_potential_test.py
@@ -4,9 +4,9 @@ from symplyphysics import (
     errors,
     units,
     Quantity,
-    SI,
     convert_to,
     assert_approx,
+    equivalent_dims,
 )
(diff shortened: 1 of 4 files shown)


test/chemistry/potential_energy_models/lennard_jones_potential_test.py
```diff
@@ -26,7 +26,7 @@ def test_args_fixture():
 
 def test_basic_law(test_args):
     result = lennard_jones_potential.calculate_potential(test_args.e, test_args.sigma, test_args.r)
-    assert SI.get_dimension_system().equivalent_dims(result.dimension, units.energy)
+    assert equivalent_dims(result, units.energy)
```
blackyblack, line 29: Revert this change. We will introduce it later.

blackyblack (approved):

test/chemistry/potential_energy_models/lennard_jones_potential_test.py
alesanter: Oh I see your new functionality, okay

PR #365, opened 2024-02-06
Law: angular momentum of rigid body

alesanter:


symplyphysics/laws/kinematic/angular_momentum_is_rotational_inertia_times_angular_velocity.py
```diff
@@ -0,0 +1,40 @@
...
+    Quantity,
+    Symbol,
+    print_expression,
+    validate_input,
+    validate_output,
+    angle_type,
+)
+
+# Description
+## For a rigid body rotating about a fixed axis, the component of its angular momentum parallel
```
blackyblack, line 13: I think angular momentum vector is always parallel with rotational axis, so basically you are calculating magnitude of angular momentum vector.

blackyblack (approved):

symplyphysics/laws/kinematic/angular_momentum_is_rotational_inertia_times_angular_velocity.py
alesanter: They are parallel as long as the axis is fixed, otherwise they need not be parallel

PR #382, opened 2024-02-08
orbital_velocity_from_length_of_large_half_axis_and_planet_mass

Darya2909:


symplyphysics/laws/gravity/orbital_velocity_from_length_of_large_half_axis_and_planet_mass.py
```diff
@@ -0,0 +1,37 @@
...
+    validate_output)
+
+## Description
+## The orbital velocity of a body (usually a planet, a natural or artificial satellite, or a multiple star) is the speed at which it rotates around the barycenter of the system, usually around a more massive body.
+## Law: V = √(G * M * ((2 / r) - 1 / a))
+## Where:
+## V is orbital velocity
+## M is mass of planet
+## G is gravitational constant
+## r is the distance between the rotating body and the central body
```
blackyblack, line 13: It seems, `r` should be <= `a`. But from mathematical point of view, formula works until `r` < `2a`. Could you clarify, which is a valid condition for this formula? Also might be a good idea to throw on invalid distance parameter.

blackyblack (approved):

blackyblack (approved):

PR #385, opened 2024-02-09
Example: measuring acceleration due to gravity

alesanter:


examples/dynamics/measuring_acceleration_due_to_gravity.py
```diff
@@ -0,0 +1,45 @@
...
+    rod_inertia_law.length, rod_length
+)
+
+rod_rotational_inertia_through_pivot = parallel_axis_theorem.law.rhs.subs({
+    parallel_axis_theorem.rotational_inertia_through_com: rod_rotational_inertia_through_com,
+    parallel_axis_theorem.mass: rod_inertia_law.mass,
+    parallel_axis_theorem.distance_between_axes: distance_to_pivot,
+})
+
+acceleration_due_to_gravity = solve(
```
blackyblack, line 34: Do you have any real numbers to put into the result and confirm it is close to actual `g`?

blackyblack (approved):

examples/dynamics/measuring_acceleration_due_to_gravity.py
alesanter: I've found some data for a simple pendulum experiment only

PR #393, opened 2024-02-10
Create: base class for laws

Voldoned:


blackyblack:
Not generalized enough.

PR #396, opened 2024-02-11
van_der_waals_equation

Darya2909:


symplyphysics/laws/thermodynamics/van_der_waals_equation.py
```diff
@@ -0,0 +1,46 @@
+from sympy import (Eq, solve)
```
blackyblack, line 1: Already exists: https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/thermodynamics/equations_of_state/van_der_waals_state_equation.py
I suggest to ask me before taking new law.

blackyblack (changes requested):

PR #403, opened 2024-02-19
Commit #1. general_definition_of_entropy

Zulusssss:
Add law and tests.


symplyphysics/laws/thermodynamics/general_definition_of_entropy.py
```diff
@@ -0,0 +1,29 @@
+from sympy import (Eq, solve, log)
```
blackyblack, line 1: Rename law. Entropy from something.

blackyblack (approved):

PR #405, opened 2024-02-20
Patch: QuantityVector.from_base_vector function

alesanter:


symplyphysics/core/vectors/vectors.py
```diff
@@ -143,5 +144,9 @@ def to_base_vector(self) -> Vector:
 
     @staticmethod
     def from_base_vector(vector: Vector, *,
-        dimension: Optional[Dimension] = None) -> QuantityVector:
-        return QuantityVector(vector.components, vector.coordinate_system, dimension=dimension)
+        dimension: Optional[Dimension] = None,
+        subs: Optional[dict[Expr, Quantity]] = None) -> QuantityVector:
+        components = vector.components
+        if subs is not None:
```
blackyblack, line 150: Use `if else` one-liner.

blackyblack (approved):

symplyphysics/core/vectors/vectors.py
alesanter: I would argue that would hinder readability, but on the other hand I'm fine with it

blackyblack (approved):

PR #408, opened 2024-02-20
Definition: damped oscillator equation

alesanter:


symplyphysics/definitions/damped_harmonic_oscillator_equation.py
```diff
@@ -0,0 +1,139 @@
+from sympy import (
```
blackyblack, line 1: Does not look like definition. Move it to kinematic.

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
```diff
@@ -0,0 +1,139 @@
...
+# not specifying the dimensions because it can be any physical quantity
+displacement = symbols("displacement", cls=SymFunction, real=True)
+time = Symbol("time", units.time, positive=True)
+undamped_angular_frequency = Symbol(
+    "undamped_angular_frequency", angle_type / units.time, positive=True
+)
+damping_ratio = Symbol("damping_ratio", dimensionless, positive=True)
+
+definition = (
+    Derivative(displacement(time), time, time)
```
blackyblack, line 42: Use `(time, 2)`

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
```diff
@@ -0,0 +1,139 @@
...
+    undamped_angular_frequency_: Quantity,
+    damping_ratio_: float,
+    time_: Quantity,
+) -> Quantity:
+    result = critical_displacement.subs({
+        critical_initial_position: initial_position_,
+        critical_initial_velocity: initial_velocity_,
+        undamped_angular_frequency: undamped_angular_frequency_,
+        damping_ratio: damping_ratio_,
+    }).subs(time, time_)
```
blackyblack, line 138: Why time is substituted separately?

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
```diff
@@ -0,0 +1,139 @@
...
+## x(t) = exp(-w*z*t) * (C1 * exp(w*sqrt(z**2 - 1)*t) + C2 * exp(-w*sqrt(z**2 - 1)*t))
+
+# z > 1, overdamped system
+## the system exponentially decays to steady state without oscillations
+## larger values of the damping ratio return to equilibrium more slowly
+
+overdamped_displacement = dsolve(definition, displacement(time)).rhs
+
+# z = 1, critically damped system
+## the system returns to equilibrium as quickly as possiblewithout oscillating
```
blackyblack, line 57:
```suggestion
## the system returns to equilibrium as quickly as possible without oscillating
```

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
```diff
@@ -0,0 +1,139 @@
...
+_critical_subs = (
+    definition
+    .subs(damping_ratio, 1)
+    .subs(displacement(time), critical_displacement)
+    .doit()
+    .simplify()
+)
+assert expr_equals(_critical_subs, 0)
+
+# z < 1, underdamped
```
blackyblack, line 85: You have 3 different cases. Separate them into different laws.

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
```diff
@@ -0,0 +1,139 @@
...
+)
+
+# The general solution looks like this:
+## x(t) = exp(-w*z*t) * (C1 * exp(w*sqrt(z**2 - 1)*t) + C2 * exp(-w*sqrt(z**2 - 1)*t))
+
+# z > 1, overdamped system
+## the system exponentially decays to steady state without oscillations
+## larger values of the damping ratio return to equilibrium more slowly
+
+overdamped_displacement = dsolve(definition, displacement(time)).rhs
```
blackyblack, line 54: This is general solution, not for overdamped case, if I'm getting it right.

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
```diff
@@ -0,0 +1,139 @@
...
+## assuming the body only moves in one direction.
+
+# Definition: d**2(x(t))/dt**2 + 2*zeta*omega*d(x(t))/dt + (omega**2)*x(t) = 0
+## x(t) - position of the oscillating body
+## t - time
+## omega - undamped angular frequency
+## zeta - damping ratio, the value of which critically determines the behavior of the system
+
+# not specifying the dimensions because it can be any physical quantity
+displacement = symbols("displacement", cls=SymFunction, real=True)
```
blackyblack, line 34: I think you can use `length` dimension here. It is suitable for most cases and when we'd need more general equation, we'll refactor it.

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
alesanter: At first I wanted to model it alongside `definitions/harmonic_oscillator_is_second_derivative_equation` so I put it here

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
alesanter: But the overdamped case looks exactly the same as the general solution. $`z > 1`$ doesn't bring that many interesting properties to the equation that make it stand out somehow

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
alesanter: I agree

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
blackyblack: I see. Well we can leave general equation in definitions, I guess.

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
blackyblack: Hm, you are right. Probably we should simply put it into the description of general solution and do not extract to separate law.

alesanter: I actually think it might be better to create a directory `laws/kinematics/damped_oscillations` and move the related laws there, and put the general solution there too

blackyblack: I suggest `oscillators` folder.

PR #413, opened 2024-02-22
Law: exponential decay of damped oscillator

alesanter:


symplyphysics/laws/kinematic/damped_oscillations/exponential_decay_constant.py
```diff
@@ -0,0 +1,47 @@
...
+    print_expression,
+)
+
+# Description
+## For any (positive) value of the damping ratio, the position of a damped oscillator
+## is subject to exponential decay, i.e. its position is proportional to exp(-lambda*t)
+## where lambda is the exponential decay constant and t is time.
+
+# Law: lambda = omega * zeta
+## lambda - exponential decay constant
```
blackyblack, line 19: It is not really a constant. Maybe `decay rate` is better?

symplyphysics/laws/kinematic/damped_oscillations/exponential_decay_constant.py
```diff
@@ -0,0 +1,47 @@
...
+    validate_output,
+    print_expression,
+)
+
+# Description
+## For any (positive) value of the damping ratio, the position of a damped oscillator
+## is subject to exponential decay, i.e. its position is proportional to exp(-lambda*t)
+## where lambda is the exponential decay constant and t is time.
+
+# Law: lambda = omega * zeta
```
blackyblack, line 18: Maybe it is better to write it in form `zeta = lambda / omega`. I think it is more usual to know decay rate and calculate damping ratio from it.

blackyblack, line 18: It seems, it is approximate and for underdamped case.

alesanter: I'll look into it

symplyphysics/laws/kinematic/damped_oscillations/exponential_decay_constant.py
alesanter: image
It's not a constant, but the terminology seems to be like that

symplyphysics/laws/kinematic/damped_oscillations/exponential_decay_constant.py
alesanter: I guess you're right, it makes more sense practically

symplyphysics/laws/kinematic/damped_oscillations/exponential_decay_constant.py
blackyblack: ok

symplyphysics/laws/kinematic/damped_oscillations/exponential_decay_constant.py
blackyblack: What about this?

alesanter: Why do you think it is approximate? I don't fully understand what you mean by that word. The solution of the damped oscillator equation contain the term $`\exp(-\lambda t)`$, where $`\lambda = \zeta \omega_0`$ by design of that equation.

blackyblack: My mistake. I've seen someone mentioned it, but it seems I got it wrong.

blackyblack (approved):

PR #418, opened 2024-02-23
Law: underdamped oscillator

alesanter:


symplyphysics/laws/kinematic/damped_oscillations/underdamping.py
```diff
@@ -0,0 +1,79 @@
...
+
+# Description
+## In the presence of a damping force in the oscillating system, the system's behaviour
+## depends on the value of the damping ratio. When it is less than 1, the system oscillates
+## with a slightly different frequency than in the undamped case, and its amplitude decreasing
+## to zero. This behaviour is also known as underdamping.
+
+# Law: x(t) = x_max * exp(-lambda*t) * cos(omega_damped*t + phi)
+## x(t) - position of damped oscillator
+## x_max - amplitude of damped oscillations
```
blackyblack, line 24: But amplitude changes over time. Maybe it should be initial amplitude or initial position?

symplyphysics/laws/kinematic/damped_oscillations/underdamping.py
```diff
@@ -0,0 +1,79 @@
...
+## phi - phase lag
+
+# Conditions
+## - System is underdamped, i.e. its damping ratio is less than 1.
+
+displacement = Function("displacement", units.length, real=True)
+time = Symbol("time", units.time, nonnegative=True)
+amplitude = Symbol("amplitude", units.length, positive=True)
+exponential_decay_constant = Symbol("exponential_decay_constant", 1 / units.time, positive=True)
+undamped_angular_frequency = Symbol("undamped_angular_frequency", angle_type / units.time, positive=True)
```
blackyblack, line 36: In description it is called `damped`.

symplyphysics/laws/kinematic/damped_oscillations/underdamping.py
alesanter: Hm, in reality it's none of those, but just some constant $`A`$ that needs to be found from the boundary equation $`x(0) = A \cos(\phi)`$. It would be the initial amplitude if $\phi = 0$.

symplyphysics/laws/kinematic/damped_oscillations/underdamping.py
alesanter: I'll fix it

symplyphysics/laws/kinematic/damped_oscillations/underdamping.py
blackyblack: Yes, but position is never higher than `A` for underdamped oscillator. So it is maximum amplitude.

alesanter: It is a non-strict upper boundary of $`x`$ but that doesn't mean it ever reaches that value (in the general case). Maybe we can leave it for the end user to solve

symplyphysics/laws/kinematic/damped_oscillations/underdamping.py
```diff
@@ -0,0 +1,81 @@
...
+)
+@validate_output(displacement)
+def calculate_displacement(
+    initial_position_: Quantity,
+    exponential_decay_constant_: Quantity,
+    damped_angular_frequency_: Quantity,
+    phase_lag_: Quantity | float,
+    time_: Quantity,
+) -> Quantity:
+    coefficient_expr = solve(
```
alesanter, line 70: I wonder if we should also check here for the fact that the damping ratio is indeed less than one, and otherwise throw a ValueError

alesanter: Judging by the calculations I've just made there's no need in that

blackyblack (approved):

PR #426, opened 2024-02-25
Commit #1. relative_humidity

Zulusssss:
Add law and tests.

### DIFF
--- a/symplyphysics/laws/thermodynamics/relative_humidity.py
+++ b/symplyphysics/laws/thermodynamics/relative_humidity.py
@@ -13,11 +13,16 @@ from symplyphysics import (
 ## Relative humidity is the ratio of water vapor pressure p to saturated vapor pressure p0 at a given temperature.
 ## Saturated vapor is vapor in dynamic equilibrium with a liquid or solid of the same composition in a closed system.
 ## In such a pair, the number of evaporating molecules is equal to the number of condensing molecules per unit of time.
-## Water vapor is the gaseous aggregate state of water.
+## Water vapor is the gaseous aggregate state of water. This is the vapor for which humidity is measured.
+## Pressures on both vapors should be measured the same way, at the same temperature.
+## There is also an oversaturated vapor. Supersaturated vapor is vapor whose pressure exceeds the pressure of saturated
+## vapor at a given temperature. It can be obtained by increasing the vapor pressure in a volume free from condensation
+## centers (dust particles, ions, small liquid droplets, etc.). Another method of production is cooling saturated vapor
+## under the same conditions.
 
-## Law is: ph = (p / p_sat) * 100, where
+## Law is: ph = p / p_sat, where
 ## ph - relative humidity,
-## p - water vapor pressure,
+## p - water vapor static pressure (it is called actual vapor pressure),
 ## p_sat - saturated vapor pressure.
 
 relative_humidity = Symbol("relative_humidity", dimensionless)
--- a/symplyphysics/laws/thermodynamics/relative_humidity.py
+++ b/symplyphysics/laws/thermodynamics/relative_humidity.py
@@ -1,4 +1,4 @@
-from sympy import (Eq, solve, S)
+from sympy import (Eq, solve)
 from symplyphysics import (
     units,
     Quantity,
@@ -6,27 +6,26 @@ from symplyphysics import (
     print_expression,
     validate_input,
     validate_output,
-    dimensionless,
-    convert_to
+    dimensionless
 )
 
 # Description
 ## Relative humidity is the ratio of water vapor pressure p to saturated vapor pressure p0 at a given temperature.
+## Saturated vapor is vapor in dynamic equilibrium with a liquid or solid of the same composition in a closed system.
+## In such a pair, the number of evaporating molecules is equal to the number of condensing molecules per unit of time.
+## Water vapor is the gaseous aggregate state of water.
 
 ## Law is: ph = (p / p_sat) * 100, where
 ## ph - relative humidity,
 ## p - water vapor pressure,
 ## p_sat - saturated vapor pressure.
 
-# Conditions:
-## - the process is isobaric-isothermal.
-
 relative_humidity = Symbol("relative_humidity", dimensionless)
 
 water_vapor_pressure = Symbol("water_vapor_pressure", units.pressure)
 saturated_vapor_pressure = Symbol("saturated_vapor_pressure", units.pressure)
 
-law = Eq(relative_humidity, (water_vapor_pressure / saturated_vapor_pressure) * 100)
+law = Eq(relative_humidity, water_vapor_pressure / saturated_vapor_pressure)
 
 
 def print_law() -> str:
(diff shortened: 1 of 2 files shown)


symplyphysics/laws/thermodynamics/relative_humidity.py
```diff
@@ -0,0 +1,46 @@
...
+# Description
+## Relative humidity is the ratio of water vapor pressure p to saturated vapor pressure p0 at a given temperature.
+
+## Law is: ph = (p / p_sat) * 100, where
+## ph - relative humidity,
+## p - water vapor pressure,
+## p_sat - saturated vapor pressure.
+
+# Conditions:
+## - the process is isobaric-isothermal.
```
blackyblack, line 22: What process?

symplyphysics/laws/thermodynamics/relative_humidity.py
```diff
@@ -0,0 +1,46 @@
...
+
+relative_humidity = Symbol("relative_humidity", dimensionless)
+
+water_vapor_pressure = Symbol("water_vapor_pressure", units.pressure)
+saturated_vapor_pressure = Symbol("saturated_vapor_pressure", units.pressure)
+
+law = Eq(relative_humidity, (water_vapor_pressure / saturated_vapor_pressure) * 100)
```
blackyblack, line 29: You are calculating percentage this way. I'm not sure we really ever need percentage instead of fraction.

symplyphysics/laws/thermodynamics/relative_humidity.py
```diff
@@ -0,0 +1,46 @@
...
+    validate_output,
+    dimensionless,
+    convert_to
+)
+
+# Description
+## Relative humidity is the ratio of water vapor pressure p to saturated vapor pressure p0 at a given temperature.
+
+## Law is: ph = (p / p_sat) * 100, where
+## ph - relative humidity,
```
blackyblack, line 17: Can it be higher than 100%?

symplyphysics/laws/thermodynamics/relative_humidity.py
```diff
@@ -0,0 +1,46 @@
...
+## p - water vapor pressure,
```
blackyblack, line 18: Explain it better. What is water vapor? What is its pressure?

symplyphysics/laws/thermodynamics/relative_humidity.py
```diff
@@ -0,0 +1,46 @@
...
+## p_sat - saturated vapor pressure.
```
blackyblack, line 19: What is saturated vapor?

symplyphysics/laws/thermodynamics/relative_humidity.py
```diff
@@ -0,0 +1,46 @@
...
+
+def print_law() -> str:
+    return print_expression(law)
+
+
+@validate_input(water_vapor_pressure_=water_vapor_pressure,
+    saturated_vapor_pressure_=saturated_vapor_pressure)
+@validate_output(relative_humidity)
+def calculate_relative_humidity(water_vapor_pressure_: Quantity,
+    saturated_vapor_pressure_: Quantity) -> float:
```
blackyblack, line 40: You can return Quantity to let users convert the result to percents automatically.

test/thermodynamics/relative_humidity_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+def test_bad_water_vapor_pressure(test_args: Args) -> None:
+    water_vapor_pressure = Quantity(1 * units.coulomb)
+    with raises(errors.UnitsError):
+        humidity_law.calculate_relative_humidity(water_vapor_pressure, test_args.saturated_vapor_pressure)
+    with raises(TypeError):
+        humidity_law.calculate_relative_humidity(100, test_args.saturated_vapor_pressure)
+
+
+def test_bad_saturated_vapor_pressure(test_args: Args) -> None:
```
blackyblack, line 36: Combine pressure cases in one test case.

symplyphysics/laws/thermodynamics/relative_humidity.py
Zulusssss: > Can it be higher than 100%?

Yes

test/thermodynamics/relative_humidity_test.py
```diff
@@ -22,20 +22,16 @@ def test_args_fixture() -> Args:
 
 def test_basic_relative_humidity(test_args: Args) -> None:
     result = humidity_law.calculate_relative_humidity(test_args.water_vapor_pressure, test_args.saturated_vapor_pressure)
-    assert_equal(result, 44.9)
+    assert_equal(result, 0.449)
```
blackyblack, line 25: You can also write `44.9 * units.percent` if you like it more.

symplyphysics/laws/thermodynamics/relative_humidity.py
```diff
@@ -1,32 +1,31 @@
...
+    dimensionless
 )
 
 # Description
 ## Relative humidity is the ratio of water vapor pressure p to saturated vapor pressure p0 at a given temperature.
+## Saturated vapor is vapor in dynamic equilibrium with a liquid or solid of the same composition in a closed system.
+## In such a pair, the number of evaporating molecules is equal to the number of condensing molecules per unit of time.
+## Water vapor is the gaseous aggregate state of water.
 
 ## Law is: ph = (p / p_sat) * 100, where
```
blackyblack, line 18: Remove `* 100` or mention percents.

symplyphysics/laws/thermodynamics/relative_humidity.py
blackyblack: Still not clear, what pressure do you mean here.
Actually I've tried to find, what pressure is meant here, and there is no clear answer to this. I think it should be the same for both saturated and actual vapor, so it can be any. But the easiest to find is static pressure, so it seems it is being used for calculations.
You can mention, that pressures on both vapors should be measured the same way, at the same temperature.

symplyphysics/laws/thermodynamics/relative_humidity.py
```diff
@@ -1,32 +1,31 @@
...
 ## ph - relative humidity,
 ## p - water vapor pressure,
```
blackyblack, line 20: It seems it is not just some random water vapor, but the vapor we are measuring humidity for. This is not really clear from your description.
Mention that relative humidity is for this vapor. And I see somewhere it is called `actual vapor pressure`.

symplyphysics/laws/thermodynamics/relative_humidity.py
```diff
@@ -1,32 +1,31 @@
...
-    dimensionless,
-    convert_to
+    dimensionless
 )
 
 # Description
 ## Relative humidity is the ratio of water vapor pressure p to saturated vapor pressure p0 at a given temperature.
+## Saturated vapor is vapor in dynamic equilibrium with a liquid or solid of the same composition in a closed system.
+## In such a pair, the number of evaporating molecules is equal to the number of condensing molecules per unit of time.
+## Water vapor is the gaseous aggregate state of water.
```
blackyblack, line 16: Of course. But the important part here is that it is this vapor, that we measure humidity of.

symplyphysics/laws/thermodynamics/relative_humidity.py
```diff
@@ -0,0 +1,45 @@
...
+    Quantity,
+    Symbol,
+    print_expression,
+    validate_input,
+    validate_output,
+    dimensionless
+)
+
+# Description
+## Relative humidity is the ratio of water vapor pressure p to saturated vapor pressure p0 at a given temperature.
```
blackyblack, line 13: Add some note if and when it can be higher than 100%.

blackyblack (approved):

PR #428, opened 2024-02-25
Law: phase velocity vector

alesanter:


symplyphysics/laws/waves/vector/phase_velocity_from_angular_velocity_and_wavevector.py
```diff
@@ -0,0 +1,74 @@
...
+
+def phase_velocity_law(wavevector_: Vector) -> Vector:
+    wavenumber_ = vector_magnitude(wavevector_)
+    return scale_vector(
+        angular_frequency / wavenumber_**2,
+        wavevector_,
+    )
+
+
+# k = |k| * e_k = (w / |v|) * (v / |v|) = w * v / |v|**2
```
blackyblack, line 36: It does not look very obvious. Might be a good idea to add a proof, eg by taking some generic `w` and `v`, putting them into `phase_velocity_law` and checking the result.

symplyphysics/laws/waves/vector/phase_velocity_from_angular_velocity_and_wavevector.py
```diff
@@ -0,0 +1,74 @@
...
+# Description
+## The phase velocity of a wave is the rate at which the wave propagates in a medium.
+## It is the velocity at which the phase of one frequency component of the wave travels.
+## The phase velocity is collinear with the wavevector.
+
+# Law: v = w * k / |k|**2
+## v - vector of phase velocity of wave
+## w - angular frequency of wave
+## k - wavevector of wave
+## |k| - wavenumber of wave, where |v| is the Euclidean norm of vector v
```
blackyblack, line 23: Is it the same parameter as in https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/definitions/angular_wavenumber_is_inverse_wavelength.py#L19?
I suggest to use the same name for both.

symplyphysics/laws/waves/vector/phase_velocity_from_angular_velocity_and_wavevector.py
```diff
@@ -0,0 +1,74 @@
...
+    scale_vector,
+    vector_magnitude,
+)
+
+# Description
+## The phase velocity of a wave is the rate at which the wave propagates in a medium.
+## It is the velocity at which the phase of one frequency component of the wave travels.
+## The phase velocity is collinear with the wavevector.
+
+# Law: v = w * k / |k|**2
```
blackyblack, line 19: It seems, you are taking the direction of `k` (by dividing k vector to k norm) and a magnitude from `w / k`. We know that `k` is inverse of wavelength, so basically we divide length by time and get velocity.
Sounds quite easy to write a proof.
Not in this PR though.

symplyphysics/laws/waves/vector/phase_velocity_from_angular_velocity_and_wavevector.py
alesanter: I'll add "angular" to both of them

symplyphysics/laws/waves/vector/phase_velocity_from_angular_velocity_and_wavevector.py
alesanter: Maybe it will be better to rewrite the law as `v = (w / |k|) (k / |k|)`, to understand easier where it comes from.

As for the proof, yes, it's basically another version of $`v_{\text{phase}} = \frac {\lambda} {T}`$ (which we don't have yet, so I'll add it later), but it makes more sense to add the proof to the scalar version of this law.

symplyphysics/laws/waves/vector/phase_velocity_from_angular_velocity_and_wavevector.py
alesanter: I'll add it below this law then

blackyblack (approved):

PR #431, opened 2024-02-26
Law: phase velocity from wavelength

alesanter:


symplyphysics/laws/waves/phase_velocity_from_wavelength_and_time_period.py
```diff
@@ -0,0 +1,44 @@
+from sympy import Eq
```
blackyblack, line 1: We already have this: https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/waves/wavelength_from_wave_speed_and_period.py

alesanter: Oh, I overlooked

PR #432, opened 2024-02-26
Wave equation

alesanter:
I'm not sure what `calculate` function we could have here though.

### DIFF
--- a/symplyphysics/definitions/wave_equation_in_one_dimension.py
+++ b/symplyphysics/definitions/wave_equation_in_one_dimension.py
@@ -8,8 +8,12 @@ from sympy import (
 from symplyphysics import (
     units,
     Symbol,
+    Quantity,
     print_expression,
+    validate_input,
 )
+from symplyphysics.core.expr_comparisons import expr_equals
+from symplyphysics.core.quantity_decorator import validate_output_same
 
 # Description
 ## The wave equation is a second-order linear partial differential equation used to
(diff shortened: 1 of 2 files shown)


symplyphysics/definitions/wave_equation_in_one_dimension.py
```diff
@@ -0,0 +1,38 @@
...
+)
+from symplyphysics import (
+    units,
+    Symbol,
+    print_expression,
+)
+
+# Description
+## The wave equation is a second-order linear partial differential equation used to
+## describe the description of waves, including standing wave fields such as mechanical
```
blackyblack, line 15: `describe the description` does not make sense.

symplyphysics/definitions/wave_equation_in_one_dimension.py
```diff
@@ -0,0 +1,38 @@
...
+## or electromagnetic waves.
+
+# Law: d**2(u(x, t))/dx**2 = (1/v**2) * d**2(u(x, t))/dt**2
+## u(x, t) - factor representing a displacement from rest situation
```
blackyblack, line 19: `displacement from rest situation` sounds weird. Try to look for some better description of this function.
Also, what's being displaced here?

symplyphysics/definitions/wave_equation_in_one_dimension.py
```diff
@@ -0,0 +1,38 @@
...
+## x - position
+## t - time
+## d**2/dx**2 - partial derivative w.r.t. position x
+## d**2/dt**2 - partial derivative w.r.t. time t
+## v - wave speed (a fixed non-negative real value)
```
blackyblack, line 24: Is it the same thing, as in https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/waves/phase_velocity_from_angular_frequency_and_wavenumber.py ? Maybe we could share the same name?

symplyphysics/definitions/wave_equation_in_one_dimension.py
```diff
@@ -0,0 +1,38 @@
...
+
+displacement = symbols("displacement", cls=SymFunction, real=True)
```
blackyblack, line 26: Maybe use `Function` and add dimension?

symplyphysics/definitions/wave_equation_in_one_dimension.py
```diff
@@ -0,0 +1,38 @@
...
+position = Symbol("position", units.length, real=True)
+time = Symbol("time", units.time, positive=True)
+wave_speed = Symbol("wave_speed", units.velocity, nonnegative=True)
+
+definition = Eq(
```
blackyblack, line 31: No tests. How do you know it works?

symplyphysics/definitions/wave_equation_in_one_dimension.py
blackyblack, line 19: Actually it does not look one-dimensional case. It looks like a parametrized two-dimensional (plus time) trajectory.

symplyphysics/definitions/wave_equation_in_one_dimension.py
blackyblack, line 31: Can we come up with some `calculate` function?

symplyphysics/definitions/wave_equation_in_one_dimension.py
alesanter: I meant 1 spatial dimension

symplyphysics/definitions/wave_equation_in_one_dimension.py
alesanter: Or we could include a general trigonometric solution in the form of $`u(x, t) = u_m \cos(x + v t + \phi)`$ and base the `calculate` function on that

symplyphysics/definitions/wave_equation_in_one_dimension.py
blackyblack: Yes, but it's about two spatial dimensions. One is argument of a function, and second is the function's output.

alesanter: It's the same displacement function as in the [harmonic oscillator equation](https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/definitions/harmonic_oscillator_is_second_derivative_equation.py). The $`u`$ function in this equation can be position, pressure, electric or magnetic field. Hence I think it's reasonable to not specify the dimension

symplyphysics/definitions/wave_equation_in_one_dimension.py
blackyblack: Hm, cool idea. But it does not look really general. We should clearly explain that this calculate function is only for a subset of possible functions.

symplyphysics/definitions/wave_equation_in_one_dimension.py
alesanter: They are the same

symplyphysics/definitions/wave_equation_in_one_dimension.py
blackyblack: Yes, indeed.

alesanter: image

Looking at the terminology it seems that they call it _one-dimensional_ since the function depends on one spatial dimension. What's more, the output of the function does not need to be another spatial dimension, it can be any physical quantity and maybe even the same spatial dimension.

blackyblack: Thanks.
We could add it to description though.

alesanter: I will

blackyblack (approved):

PR #437, opened 2024-02-27
Commit #1. phase_difference_of_oscillations_of_two_points

Zulusssss:
Add law and tests.

### DIFF
--- a/symplyphysics/laws/waves/phase_difference_of_oscillations_of_two_points.py
+++ b/symplyphysics/laws/waves/phase_difference_of_oscillations_of_two_points.py
@@ -1,4 +1,4 @@
-from sympy import (Eq, solve, pi, S)
+from sympy import (Eq, solve, pi, S, Abs)
 from symplyphysics import (
     units,
     Quantity,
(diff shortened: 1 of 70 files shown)


symplyphysics/laws/waves/phase_difference_of_oscillations_of_two_points.py
```diff
@@ -0,0 +1,48 @@
...
+    validate_output,
+    convert_to,
+    dimensionless,
+)
+
+# Description
+## The phase difference of the oscillations of two points spaced at distances r1 and r2 from the source of
+## the oscillations depends on these distances and wavelength.
+
+## Law is: phi = 2 * pi * (r2 - r1) / L, where
```
blackyblack, line 17: It seems, from choosing which is the first point, depends the sign of the result. Is there any way to choose first point, so the sign does not change arbitrarily? Or maybe it is better to use absolute value of difference of distances?

Zulusssss: > It seems, from choosing which is the first point, depends the sign of the result. Is there any way to choose first point, so the sign does not change arbitrarily? Or maybe it is better to use absolute value of difference of distances?

Changed the code - added Abs().
from sympy import Abs
law = Eq(phase_difference, 2 * pi * Abs(distance_second_point - distance_first_point) / wavelength)

Now I get the following error.
image

blackyblack: Will look into it.

PR #447, opened 2024-02-29
Law: displacement in simple harmonic motion

alesanter:


symplyphysics/laws/kinematic/displacement_in_simple_harmonic_motion.py
```diff
@@ -0,0 +1,87 @@
...
+_A, _B = symbols("A B", real=True)
+
+_dsolved = dsolve(_eqn, displacement(time)).rhs.subs({
+    "C1": _A,
+    "C2": _B,
+})
+
+# Setting amplitude equal to sqrt(A**2 + B**2) we can have these identities
+_eqn_A = Eq(_A/amplitude, -1 * sin(phase_lag))
+_eqn_B = Eq(_B/amplitude, cos(phase_lag))
```
blackyblack, line 54: I do not see, where does it come from.

alesanter: I think I found a way to avoid this

blackyblack (approved):

PR #451, opened 2024-03-01
Example: simple harmonic oscillations of spring

alesanter:


examples/kinematics/simple_harmonic_oscillations_of_spring.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env python3
+
+from sympy import solve, symbols, Eq
+from symplyphysics import Quantity, units, print_expression, convert_to
+from symplyphysics.laws.dynamics import (
```
blackyblack, line 5: It uses laws from `dynamics` so probably should be in `dynamics` as well.

examples/kinematics/simple_harmonic_oscillations_of_spring.py
```diff
@@ -0,0 +1,89 @@
...
+    harmonic_law.phase_lag: phase_lag,
+    harmonic_law.time: time,
+})
+
+velocity_expr = position_expr.diff(time)
+
+# (a) Amplitude of oscillations
+
+amplitude_eqn = Eq(
+    position1 ** 2 + (velocity1 / angular_frequency_expr)**2,
```
blackyblack, line 57: Could you justify it?

examples/kinematics/simple_harmonic_oscillations_of_spring.py
alesanter: Hm, that's sensible, I'll move it

examples/kinematics/simple_harmonic_oscillations_of_spring.py
alesanter: The reason I'm doing this is that technically I could just solve this and get the correct formulas:
image
But the solution `sympy` produces is pretty unattractive and it doesn't make sense from first sight:
image
I tried to find a way for `sympy` to simplify the output but no luck

examples/kinematics/simple_harmonic_oscillations_of_spring.py
blackyblack: Sync with repository before doing it. I've added some improvements.

blackyblack (approved):

PR #464, opened 2024-03-07
Law: position of antinodes in standing wave

alesanter:


symplyphysics/laws/waves/position_of_antinodes_in_standing_wave.py
```diff
@@ -0,0 +1,70 @@
...
+# Law: x_antinode = (m + 1/2) * (lambda / 2)
+## x_antinode - position of m-th antinode
+## m - integer
+## lambda - wavelength
+
+antinode_position = Symbol("antinode_position", units.length, real=True)
+integer_factor = Symbol("integer_factor", dimensionless, integer=True)
+wavelength = Symbol("wavelength", units.length, positive=True)
+
+law = Eq(antinode_position, (integer_factor + S(1)/2) * wavelength / 2)
```
blackyblack, line 29: `1/2` does not work here?

symplyphysics/laws/waves/position_of_antinodes_in_standing_wave.py
```diff
@@ -0,0 +1,70 @@
...
+
+_standing_wave_spatial_derivative_at_antinodes = _standing_wave_spatial_derivative.subs(
+    standing_wave_law.position, law.rhs
+)
+
+assert expr_equals(_standing_wave_spatial_derivative_at_antinodes, 0)
+
+# Since the spatial derivative of the wave is zero in antinodes, according to calculus
+# these are the points of extrema of the wave, therefore the locations of the maximum
+# amplitude of the wave.
```
blackyblack, line 50: Why not minimal amplitude?

symplyphysics/laws/waves/position_of_antinodes_in_standing_wave.py
alesanter: `1/2` would yield a `float` which is approximate but `S(1)/2` is exact. I think we have to keep fractions exact in formulas

symplyphysics/laws/waves/position_of_antinodes_in_standing_wave.py
alesanter: image
Judging by the definition of amplitude, _maximum amplitude_ would refer to both points of maximum and minimum of the wave since amplitude is always positive

blackyblack: Ok. Add this explanation to the proof.

symplyphysics/laws/waves/position_of_antinodes_in_standing_wave.py
blackyblack: Ok. Maybe `S.One` looks a bit better.

symplyphysics/laws/waves/position_of_antinodes_in_standing_wave.py
alesanter: Okay

symplyphysics/laws/waves/position_of_antinodes_in_standing_wave.py
alesanter: I agree

blackyblack (approved):

PR #480, opened 2024-03-11
Prove: general solution of 1D wave equation

alesanter:


symplyphysics/laws/waves/wave_equation_general_solution_in_one_dimension.py
```diff
@@ -32,7 +37,37 @@
...
+
+_eqn = wave_eqn.definition.subs({
+    wave_eqn.position: position,
+    wave_eqn.time: time,
+    wave_eqn.phase_velocity: _phase_velocity,
+})
+
+_eqn_subs = _eqn.subs(
+    wave_eqn.displacement(position, time), _solution
+).doit()
```
blackyblack, line 68: I'm not sure, what's exactly proven here. I see that before `doit()` there is some equation with second order derivatives of `solution_function` over `dx2` and `dt2`. But I cannot understand, what it is about.
Also, having `True` as a result of `doit()` might happen when SymPy cannot properly resolve the equation.

alesanter: I aimed to show that the solution function presented in this law satisfies the wave differential equation, so I did a simple substitution and made `sympy` do the differentiations with `.doit()`. So there's no solving here involved. It seems to me that the `True` in the end is actually `sympy`'s way of dealing with equalities where both parts are equal, e.g. plugging `Eq(5, 5)` or `Eq(x, x)` yield `True` in `sympy`.

alesanter: Also I've just seen there's `sympy.solvers.pde.pdsolve`, but it doesn't seem to work for the wave equation

alesanter: Although it does seem to work for a one-way wave equation, namely $`\frac{\partial u}{\partial t} \pm v \frac{\partial u}{\partial x} = 0`$, if we want to solve the differential equation directly and then compare the result to what is presented in the law.

blackyblack (approved):

PR #518, opened 2024-03-20
Derive: standing wave

alesanter:


symplyphysics/laws/waves/displacement_in_standing_wave.py
```diff
@@ -37,7 +39,15 @@
 law = Eq(displacement(position, time),
     (2 * amplitude) * sin(angular_wavenumber * position) * cos(angular_frequency * time))
 
-# TODO: Derive from the sum of two traveling waves
+# Derive from the sum of two traveling waves
+
+_forward_wave = amplitude * sin(angular_wavenumber * position - angular_frequency * time)
```
blackyblack, line 44: Mention that you've taken the wave form from the notes above.
Also might be a good idea to use https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/waves/phase_of_traveling_wave.py to justify the wave function phase.

blackyblack (approved):

symplyphysics/laws/waves/displacement_in_standing_wave.py
alesanter: I agree

blackyblack (approved):

PR #524, opened 2024-03-20
Law: Maxwell—Boltzmann speed distribution

alesanter:


symplyphysics/laws/thermodynamics/maxwell_boltzmann_distributions/speed_distribution.py
```diff
@@ -0,0 +1,62 @@
...
+## T - equilibrium temperature of the particle ensemble
+
+# Conditions
+## - Number of particles is big enough that the laws of thermodynamics can be applied.
+## - Particles are identical, non-interacting, non-relativistic, and obeying classical laws of physics.
+## - The ensemble of particles is at thermodynamic equilibrium.
+
+speed_distribution_function = Symbol("speed_distribution_function", 1 / units.velocity)
+particle_speed = Symbol("particle_speed", units.velocity)
+particle_mass = Symbol("particle_mass", units.mass)
```
blackyblack, line 30: We are introducing global symbols. Take a look here: https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/gravity/gravitational_potential_energy.py#L27
And define `particle_mass` with `clone_symbol()`

test/thermodynamics/maxwell_boltzmann_distributions/speed_distribution_test.py
```diff
@@ -0,0 +1,55 @@
...
+    Quantity,
+)
+from symplyphysics.core.symbols.celsius import Celsius, to_kelvin_quantity
+from symplyphysics.laws.thermodynamics.maxwell_boltzmann_distributions import (
+    speed_distribution as distribution_law,
+)
+
+# Description
+## The value of the speed distribution function for an ensemble of Argon particles (particle mass m = 39.948 u)
+## at equilibrium temperature T = 10°C at speed v = 10 m/s is 5.57e-6 s/m.
```
blackyblack, line 16: `seconds / meter` is valid unit of measurement, but it reflects amount of particles per unit speed, so I think it is better to use `1 / (meter / second)`

symplyphysics/laws/thermodynamics/maxwell_boltzmann_distributions/speed_distribution.py
alesanter: Okay. Why not make it a method, say `.clone(new_name)`, rather than a function?

test/thermodynamics/maxwell_boltzmann_distributions/speed_distribution_test.py
alesanter: Sure

symplyphysics/laws/thermodynamics/maxwell_boltzmann_distributions/speed_distribution.py
blackyblack: I prefer free functions over class methods. Maybe will refactor this way if we find it more suitable.

alesanter: I see. As far as I saw `sympy` offers both most of the time

PR #525, opened 2024-03-20
Definition: radiant exitance of a surface

alesanter:


symplyphysics/definitions/radiant_exitance_is_radiant_flux_emitted_per_unit_area.py
```diff
@@ -0,0 +1,47 @@
...
+@validate_input(
+    radiant_flux_=radiant_flux,
+    surface_area_=surface_area,
+)
+@validate_output(radiant_exitance)
+def calculate_radiant_exitance(
+    radiant_flux_: Quantity,
+    surface_area_: Quantity,
+) -> Quantity:
+    """\
```
blackyblack, line 40: You do not need `\` symbols here.
Also regular Python comments look better.

symplyphysics/definitions/radiant_exitance_is_radiant_flux_emitted_per_unit_area.py
```diff
@@ -0,0 +1,47 @@
...
+    Calculate radiant exitance for a surface small enough that \
```
blackyblack, line 41: You simply use constant flux function. It does not really matter if surface is small or not.

blackyblack (approved):

symplyphysics/definitions/radiant_exitance_is_radiant_flux_emitted_per_unit_area.py
alesanter: Hm, yeah, I agree

blackyblack (approved):

PR #530, opened 2024-03-21
Derive: average square of speed in Maxwell-Boltzmann statistics

alesanter:


symplyphysics/laws/thermodynamics/maxwell_boltzmann_distributions/speed_distribution.py
```diff
@@ -27,10 +27,10 @@
 ## - Particles are identical, non-interacting, non-relativistic, and obeying classical laws of physics.
 ## - The ensemble of particles is at thermodynamic equilibrium.
 
-speed_distribution_function = Symbol("speed_distribution_function", 1 / units.velocity)
-particle_speed = Symbol("particle_speed", units.velocity)
-particle_mass = clone_symbol(symbols.basic.mass, "particle_mass")
-equilibrium_temperature = clone_symbol(symbols.thermodynamics.temperature, "equilibrium_temperature")
+speed_distribution_function = Symbol("speed_distribution_function", 1 / units.velocity, positive=True)
```
blackyblack, line 30: Btw, shouldn't it be a function?

symplyphysics/laws/thermodynamics/maxwell_boltzmann_distributions/speed_distribution.py
```diff
@@ -27,10 +27,10 @@
...
+particle_speed = Symbol("particle_speed", units.velocity, nonnegative=True)
```
blackyblack, line 31: Maybe use `positive` instead?

symplyphysics/laws/thermodynamics/average_square_of_velocity.py
```diff
@@ -1,22 +1,39 @@
...
 law = Eq(average_square_velocity, 3 * units.boltzmann_constant * temperature_in_gas / molecule_mass)
 
+# Derive from the Maxwell-Boltzmann speed distribution
+
+_distribution = speed_distribution.law.rhs.subs({
+    speed_distribution.particle_mass: molecule_mass,
+    speed_distribution.equilibrium_temperature: temperature_in_gas,
+})
+
+_average_square_of_speed_derived = integrate(
```
blackyblack, line 30: Elaborate a bit on this formula. Why do you get average square velocity as a result?

symplyphysics/laws/thermodynamics/maxwell_boltzmann_distributions/speed_distribution.py
alesanter: Yeah, it probably should be

symplyphysics/laws/thermodynamics/maxwell_boltzmann_distributions/speed_distribution.py
alesanter: But it can be 0

symplyphysics/laws/thermodynamics/average_square_of_velocity.py
alesanter: Will do

symplyphysics/laws/thermodynamics/maxwell_boltzmann_distributions/speed_distribution.py
blackyblack: Temperature can also be 0.

alesanter: Then it makes sense to mark temperature as `nonnegative` as well. But the number of cases where we have to deal with zero absolute temperature is, in my opinion, much smaller than those with zero speed

blackyblack: I want these assumptions to be consistent all over the code. So if we use `nonnegative` parameter, every `positive` should be updated to it. There are many.

alesanter: Well now I agree that in this very formula it's more rational to have `positive` since the RHS is positive (unless `T` is zero but that's rare). Btw most of the codebase (95%-ish) doesn't use any assumptions at all (and I guess we should add them)

blackyblack: You can take on this task actually - start adding assumptions all over the code, including fixing the `positive` assumption. It's a big, but relatively easy task.

alesanter: The reason I add them is because `sympy` doesn't otherwise want to solve or simplify equations the way we need them because all the symbols are `complex` unless stated otherwise. Specifically in these laws where we're dealing with the speed distribution function.

For example, this is what happens in the law of the mean square of speed:
image

Which immediately simplifies to the following once the correct assumptions are added:
image

blackyblack: Yes, I understand it. But for the sake of making proof work, we could add assumptions as soon as we are trying to prove it. When we start adding assumption right in the laws, we are making them a part of those laws. Now we should look carefully, if law makes sense for some negative parameters, for example.

blackyblack: For physcial laws I expect most variables to be `real` and `nonnegative`.

PR #538, opened 2024-03-22
Law: discrete Maxwell—Boltzmann distribution

alesanter:


symplyphysics/laws/thermodynamics/maxwell_boltzmann_statistics/discrete_distribution.py
```diff
@@ -0,0 +1,69 @@
...
+    symbols,
+    clone_symbol,
+)
+
+# Description
+## Maxwell-Boltzmann distribution can be written as a discrete distribution of a single particle's
+## discrete energy spectrum. Maxwell-Boltzmann statistics gives the average number of particles
+## found in a given single-particle microstate.
+
+# Law: N_i / N = exp(-E_i / (k * T)) / Z
```
alesanter, line 19: If #539 is merged, we could have Boltzmann factor `f_i` here instead of `exp(-E_i / (k * T))`

blackyblack: Laws are weekly coupled, so you can use any law form you find suitable, disregard merged PRs. But if you write it as `exp(-E_i / (k * T))` and it's a copypaste of existing law, you should link them together. We usually do it by deriving one law from another.

blackyblack (approved):

PR #543, opened 2024-03-22
Law: entropy increase in reversible process

alesanter:


symplyphysics/laws/thermodynamics/entropy_increment_in_reversible_process.py
```diff
@@ -0,0 +1,53 @@
...
+## δQ - infinitesimal amount of heat transfered to system of interest (B)
+## T - thermodynamic temperature of the systems (A and B)
+
+# Notes
+## - dS is an exact differential (since entropy S is a state function) whereas δQ is an inexact differential
+## - Also applicable to actually possible quasistatic irreversible processes without composition change
+
+infinitesimal_entropy_change = Symbol("infinitesimal_entropy_change", units.energy / units.temperature)
+infinitesimal_transfer_of_heat = Symbol("infinitesimal_transfer_of_heat", units.energy)
+thermodynamic_temperature = clone_symbol(symbols.thermodynamics.temperature, "thermodynamic_temperature")
```
blackyblack, line 31: Why not just `temperature`?
Also might be a good idea to indicate that it is a summary temperature of both systems: `total_temperature`

blackyblack (approved):

symplyphysics/laws/thermodynamics/entropy_increment_in_reversible_process.py
alesanter: To me `total` looks like we're summing up the temperatures rather than saying this is the temperature shared by both systems. Maybe `common_temperature` will be better.

As for the wording of `thermodynamic_temperature`, this is how it is usually presented in most sources on statistical physics, but it's the same as `absolute_temperature` and is opposed some empirical temperature measured by a thermometer. But of course these are all the same.

blackyblack: For me `thermodynamic_temperature` does not make sense, since temperature is thermodynamic characteristic. `common_temperature` sounds good.

blackyblack (approved):

PR #547, opened 2024-03-23
Definition: volumetric coefficient of thermal expansion

alesanter:


symplyphysics/definitions/volumetric_coefficient_of_thermal_expansion.py
```diff
@@ -0,0 +1,64 @@
...
+    symbols,
+)
+
+# Description
+## The coefficient of thermal expansion describes how the size of an object changes with a change in temperature
+## at constant pressure.
+
+# Law: alpha_V = (dV/dT)_p / V
+## alpha_V - volumetric coefficient of thermal expansion
+## V - volume of body (gas, liquid or solid)
```
blackyblack, line 19: We either should take volume at some temperature point, or we'll result in coefficient as a function.

symplyphysics/definitions/volumetric_coefficient_of_thermal_expansion.py
```diff
@@ -0,0 +1,64 @@
...
+    volume_function = (
+        volume_before_
+        + (volume_after_ - volume_before_)
+        * (temperature - temperature_before_)
+        / (temperature_after_ - temperature_before_)
+    )
+    result = law.rhs.subs(
+        volume(temperature), volume_function
+    ).doit().subs(
+        temperature, (temperature_after_ + temperature_before_) / 2,
```
blackyblack, line 62: Why did you choose average temperature to calculate the coefficient?

test/definitions/volumetric_coefficient_of_thermal_expansion_test.py
```diff
@@ -0,0 +1,53 @@
...
+    assert_equal,
+    errors,
+    units,
+    Quantity,
+)
+from symplyphysics.definitions import volumetric_coefficient_of_thermal_expansion as coefficient_def
+
+# Description
+## A body is heated up and changes its volume from 1.00 m**3 at temperature T = 400 K to 1.001 m**3 at
+## T = 401 K. Then its volumetric coefficient of thermal expansion is 9.9e-3 1/K.
```
blackyblack, line 13: Coefficient at what temperature?

symplyphysics/definitions/volumetric_coefficient_of_thermal_expansion.py
alesanter: Yeah here it should be `(dV/dT)_p / V(T)`

symplyphysics/definitions/volumetric_coefficient_of_thermal_expansion.py
alesanter: I can add a comment that the volume value is calculated in the temperature middle point, but overall I'd argue that the decision to choose between the left/right/middle point is not fixed and is up to us.

blackyblack: Using `temperature_after` seems to be easiest solution.
Another option is to use volume at standard temperature as parameter for the function. This way we could reflect the idea, how the standard coefficient is calculated.

blackyblack (approved):

PR #564, opened 2024-03-27
Law: first law of thermodynamics

alesanter:


symplyphysics/laws/thermodynamics/internal_energy_change_via_amount_of_heat_and_work_done.py
```diff
@@ -0,0 +1,60 @@
...
+##   work done by the system).
+
+# Conditions
+## - This formula applies to any infinitesimal thermodynamic process, be it (quasi)static or not.
+
+internal_energy_change = Symbol("internal_energy_change", units.energy)
+heat_supplied_to_system = Symbol("heat_supplied_to_system", units.energy)
+work_done_by_system = Symbol("work_done_by_system", units.energy)
+
+law = Eq(internal_energy_change, heat_supplied_to_system - work_done_by_system)
```
blackyblack, line 40: I think that with some amount of heat supplied to the system and same amount of work done by the system, internal energy should be zero. But with your signs convention it amounts to double amount of heat supplied.

test/thermodynamics/internal_energy_change_via_amount_of_heat_and_work_done_test.py
```diff
@@ -0,0 +1,42 @@
...
+    prefixes,
+    units,
+    Quantity,
+)
+from symplyphysics.laws.thermodynamics import (
+    internal_energy_change_via_amount_of_heat_and_work_done as first_law,
+)
+
+# Description
+## The amount of heat supplied to the system is 1 J and the work done by the system is 2 J.
```
blackyblack, line 15: According to your convention, it is work done to the system.

blackyblack: Still not fixed.
Here is quote from your description:
"The sign before work `W` is negative for work done _by_ the system"
So here it should be either `-2 J` or negative sign should be applied when passed to function.

alesanter: If the work is done by the system, the sign before the "delta W" should be negative, but the actual "delta W" is positive

blackyblack: You do not need to play with description. You already have `minus` sign in formula so it should be _positive_ work done by the system. Or you can keep it negative but change the sign in formula to `plus`

alesanter: I propose making a separate law stating that `work_done_on_system = -1 * work_done_by_system` because this is a common source of confusion

blackyblack: We have similar issues like force of action and reaction. I suggest to properly describe signs and that's all.

blackyblack (approved):

PR #565, opened 2024-03-27
Law: infinitesimal increment of work in quasi-static process

alesanter:


symplyphysics/laws/thermodynamics/infinitesimal_work_in_quasistatic_process.py
```diff
@@ -0,0 +1,46 @@
...
+)
+
+# Description
+## For a process in a closed system, occuring slowly enough for it to be quasi-static (i.e. the system
+## remains in internal physical, but not necessarily chemical, thermodynamic equilibrium), the infinitesimal
+## increment of work done _by_ the system is related to the pressure inside the system and the infinitesimal
+## increment of volume of the system.
+
+# Law: delta W = P * dV
+## delta W - inexact differential of an infinitesimal increment of work done _by_ the system
```
blackyblack, line 18: `infinitesimal increment` is differential itself. Do you mean second derivative? Or it is a dublicate wording?

alesanter: Infinitesimal describes the kind of increment, because the increment itself can be finite.

alesanter: image
image
I take the wording from English-spoken sources

blackyblack: Yes but you also mention `inexact differential` which assumes infinitesimal increment. So it is a dublicate wording IMO.

alesanter: I'll move it to a note below

symplyphysics/laws/thermodynamics/infinitesimal_work_in_quasistatic_process.py
```diff
@@ -0,0 +1,49 @@
...
+# Description
+## For a process in a closed system, occuring slowly enough for it to be quasi-static (i.e. the system
+## remains in internal physical, but not necessarily chemical, thermodynamic equilibrium), the infinitesimal
+## increment of work done _by_ the system is related to the pressure inside the system and the infinitesimal
+## increment of volume of the system.
+
+# Law: delta W = P * dV
+## delta W - an infinitesimal increment of work done _by_ the system
+## P - pressure inside the system
+## dV - exact differential of an infinitesimal increment of volume of the system
```
blackyblack, line 20: Here is the same problem.
Maybe we can avoid mentioning differential and simply state that small amount of volume change results in small amount of work done?

alesanter: Oh yeah, I see now. Technically this is to be read as "infinitesimal increment which is an exact differential", but it is a double entendre in here, so I agree this has to go somewhere else. I don't think omitting the information about the type of differentials is a good idea since it is important to tell them apart in thermodynamics.

blackyblack (approved):

PR #567, opened 2024-03-27
Law: enthalpy differential

alesanter:


symplyphysics/laws/thermodynamics/enthalpy_change_via_entropy_change_and_pressure_change.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@validate_input(
+    temperature_=temperature,
+    entropy_change_=entropy_change,
+    volume_=volume,
+    pressure_change_=pressure_change,
+)
+@validate_output(enthalpy_change)
+def calculate_entropy_change(
```
blackyblack, line 59:
```suggestion
def calculate_enthalpy_change(
```

test/thermodynamics/enthalpy_change_via_entropy_change_and_pressure_change_test.py
```diff
@@ -0,0 +1,66 @@
...
+## drops by 1e-5 J/K and its pressure rises by 0.01 Pa. The change in system's enthalpy
+## amounts to 9 mJ.
+
+Args = namedtuple("Args", "t ds v dp")
+
+
+@fixture(name="test_args")
+def test_args_fixture() -> Args:
+    t = Quantity(400 * units.kelvin)
+    ds = Quantity(1e-5 * units.joule / units.kelvin)
```
blackyblack, line 26: Does not correspond to description.

symplyphysics/laws/thermodynamics/enthalpy_change_via_entropy_change_and_pressure_change.py
```diff
@@ -0,0 +1,71 @@
...
+# Note
+## - Entropy `S` and pressure `p` are so called natural variables of enthalpy as a
+##   thermodynamic potential
+
+# Conditions
+## - The system is closed and homogeneous
+## - The system is in thermal equilibrium with the environment
+## - Only reversible prossesses or pure heat transfer are considered
+
+enthalpy_change = Function("enthalpy_change", units.energy)
```
blackyblack, line 34: BTW why is this a function? I do not see it should be a function in original description. And what's point making it a function here?

alesanter: Because this differential sort of depends on the value of the other two differentials, at least that's how I imagine it in the absence of exact differentials. Ideally we should only have an enthalpy function which depends on entropy and pressure. I can also remove them and we will be left with symbols

blackyblack: I think we should not have functions unless we differentiate or integrate them in the law.

alesanter: I see

symplyphysics/laws/thermodynamics/enthalpy_change_via_entropy_change_and_pressure_change.py
```diff
@@ -31,14 +30,14 @@
 ## - The system is in thermal equilibrium with the environment
 ## - Only reversible prossesses or pure heat transfer are considered
 
-enthalpy_change = Function("enthalpy_change", units.energy)
+enthalpy_change = Symbol("enthalpy_change", units.energy)
```
blackyblack, line 33: You can refactor other laws the same way.

blackyblack (approved):

PR #588, opened 2024-04-03
Law: virial equation of state

alesanter:


symplyphysics/laws/thermodynamics/equations_of_state/virial_equation.py
```diff
@@ -0,0 +1,62 @@
...
+    validate_output,
+    convert_to,
+)
+
+# Description
+## Also called the virial expansion, the virial equation of state expresses the compressibility factor
+## (and therefore the pressure) of a real gas in local equilibrium as a power series of the density.
+
+# Law: Z = 1 + B * rho + C * rho**2 + ...
+## Z - compressibility factor; TODO: add link to definition
```
blackyblack, line 18: I think you can add it now.

symplyphysics/laws/thermodynamics/equations_of_state/virial_equation.py
```diff
@@ -0,0 +1,62 @@
...
+# Note
+## - The first virial coefficient A is defined as the constant of value 1 in order to enforce that the equation
+##   reduces to the ideal gas equation as the gas density approaches zero.
+## - N-th virial coefficient represents the non-additive n-body interactions of particles plus all
+##   mutual interactions of 2 up to `n - 1` particles.
+## - 4- and more body interactions are quite rare to happen, so the expansion is truncated to contain only
+##   the second and third virial coefficients. Moreover, the latter have been extensively studied and tabulated
+##   for many fluids.
+
+compressibility_factor = Symbol("compressibility_factor", dimensionless, positive=True)
```
blackyblack, line 31: It seems there are such coefficients that might result in negative factor. Maybe we should check for negative result and throw an exception?

alesanter: Looking at the codebase we have about 40 similar laws where we return a float and in none of them do we check for the sign of the result. In some laws we have integers as input and we don't check that they are positive either. I think it is better to leave the sign check to the user.

blackyblack: Yes, but in other laws we do not explicitly indicate that the result is positive.

symplyphysics/laws/thermodynamics/equations_of_state/virial_equation.py
```diff
@@ -59,4 +59,8 @@ def calculate_compressibility_factor(
         third_virial_coefficient: third_virial_coefficient_,
         molar_density: molar_density_,
     })
-    return float(convert_to(Quantity(result), S.One))
+
+    result_value = float(convert_to(Quantity(result), S.One))
+    if result_value < 0:
+        raise ValueError(f"Compressibility factor cannot be negative, got {result_value} instead")
```
blackyblack, line 65: We need a test for this case now.

blackyblack (approved):

PR #590, opened 2024-04-03
Law: second virial coefficient for van der Waals equation of state

alesanter:


symplyphysics/laws/thermodynamics/second_virial_coefficient_for_van_der_waals_equation_of_state.py
```diff
@@ -0,0 +1,66 @@
...
+## to inverse molar volume or equivalently to molar density.
+
+# Law: B(T) = b - a / (R * T)
+## B(T) - second virial coefficient of the virial expansion
+## a - parameter of the van der Waals equation of state representing the magnitude of intermolecular forces
+## b - parameter of the van der Waals equation of state representing the effective molecular size
+## R - molar gas constant
+## T - absolute temperature
+
+second_virial_coefficient = Function("second_virial_coefficient", units.volume / units.amount_of_substance)
```
blackyblack, line 27: I think we've agreed that function on the left side of equation does not make much sense. Symbol is just as good.

blackyblack (approved):

symplyphysics/laws/thermodynamics/second_virial_coefficient_for_van_der_waals_equation_of_state.py
alesanter: Hm, I see, though we haven't explicitly talked about that

blackyblack (approved):

PR #591, opened 2024-04-03
Law: molar internal energy of van der Waals fluid

alesanter:


symplyphysics/laws/thermodynamics/molar_internal_energy_of_van_der_waals_fluid.py
```diff
@@ -0,0 +1,73 @@
...
+## u - molar internal energy of van der Waals fluid at constant temperature
+## c_V - isochoric molar heat capacity as a function of temperature
+## T - absolute temperature
+## a - parameter of van der Waals equation of state corresponding to magnitude of intermolecular forces
+## v - molar volume
+
+# Conditions
+## - The fluid is homogeneous and in a single phase state.
+
+molar_internal_energy = Function("molar_internal_energy", units.energy / units.amount_of_substance)
```
blackyblack, line 27: Do we really need a function here?

blackyblack (approved):

symplyphysics/laws/thermodynamics/molar_internal_energy_of_van_der_waals_fluid.py
alesanter: I think yes because the user might want to do some stuff with this equation (say, solving for $`c_V`$) and then apply differentiation w.r.t. temperature.

blackyblack: But in this case function should be substituted for something differentiable. So symbol will be as good as a function. My point is we do not know in advance, what user will want to do with the formula. So putting functions here and there looks rather arbitrary. Let's use symbols unless we want to integrate or differentiate something.

alesanter: Fine, I agree with you

blackyblack (approved):

PR #593, opened 2024-04-04
Law: thermodynamical critical point

alesanter:


symplyphysics/laws/thermodynamics/critical_point_is_isotherm_stationary_inflection_point.py
```diff
@@ -0,0 +1,93 @@
...
+    symbols,
+)
+from symplyphysics.laws.thermodynamics.equations_of_state import (
+    van_der_waals_state_equation as van_der_waals_law,
+)
+
+# Description
+## Critical point (in the thermodynamic sense) is such values of volume, pressure, and temperature at 
+## which only one phase exists and at the vicinity of which the physical property of the phases of the 
+## substance change dramatically. Algebraically, the critical point is the stationary inflection point
```
blackyblack, line 18:
```suggestion
## substance changes dramatically. Algebraically, the critical point is the stationary inflection point
```

symplyphysics/laws/thermodynamics/critical_point_is_isotherm_stationary_inflection_point.py
```diff
@@ -0,0 +1,93 @@
...
+## - These equations need to be solved together with the equation of state.
+
+pressure = Function("pressure", units.pressure)
+volume = Symbol("volume", units.volume)
+temperature = symbols.thermodynamics.temperature
+
+inflection_point_eqn = Eq(Derivative(pressure(volume), volume, 1), 0)
+flat_tangent_eqn = Eq(Derivative(pressure(volume), volume, 2), 0)
+
+law = inflection_point_eqn, flat_tangent_eqn
```
blackyblack, line 37: This law looks quite weird.
We've been using this approach, but it was used as a condition to prove some other law. It is a good idea to extract conditions to other entities, but laws do not look suitable for this task.
I think we should add a separate folder for conditions. Also, there is no much sense in having `calculate` functions for them. Also not sure if we need tests for them.

symplyphysics/laws/thermodynamics/critical_point_is_isotherm_stationary_inflection_point.py
```diff
@@ -0,0 +1,93 @@
...
+
+# Find critical point for van der Waals equation of state
```
blackyblack, line 39: Irrelevant for this law. You better move this code into the critical point law.

symplyphysics/laws/thermodynamics/critical_point_is_isotherm_stationary_inflection_point.py
blackyblack: Yes. For example, if we separate critical point law to 3 laws, we now have a law for critical pressure. We want to derive it from van der Waals equation. We import critical point conditions and derive the critical point like you did below.

symplyphysics/conditions/thermodynamics/critical_point_is_isotherm_stationary_inflection_point.py
```diff
@@ -0,0 +1,31 @@
+from sympy import Eq, Derivative
```
blackyblack, line 1: Put empty `__init__.py` file in each of the new folders.

blackyblack, line 1: Do you have a better name for this condition?

symplyphysics/conditions/thermodynamics/critical_point_is_isotherm_stationary_inflection_point.py
```diff
@@ -0,0 +1,31 @@
...
+## V - volume
+## T - temperature
+## (d/dV)_T - partial derivative with respect to volume at constant temperature
+
+# Note
+## - These equations need to be solved together with the equation of state.
+
+pressure = Function("pressure", units.pressure)
+volume = Symbol("volume", units.volume)
+temperature = symbols.thermodynamics.temperature
```
blackyblack, line 26: Temperature is unused.

symplyphysics/conditions/thermodynamics/critical_point_is_isotherm_stationary_inflection_point.py
```diff
@@ -0,0 +1,31 @@
...
+
+inflection_point_condition = Eq(Derivative(pressure(volume), volume, 1), 0)
+flat_tangent_condition = Eq(Derivative(pressure(volume), volume, 2), 0)
+
+conditions = inflection_point_condition, flat_tangent_condition
```
blackyblack, line 31: Let's make it a list: `conditions = [inflection_point_condition, flat_tangent_condition]`

symplyphysics/conditions/thermodynamics/critical_point_is_isotherm_stationary_inflection_point.py
alesanter: The name reflects that for something to be a critical point it needs to be both a stationary and an inflection point of an isotherm. Is there any guidelines you have regarding naming the conditions files?

PR #596, opened 2024-04-04
Law: molar mass via particle mass

alesanter:


symplyphysics/laws/chemistry/molar_mass_via_mass_of_single_particle.py
```diff
@@ -0,0 +1,53 @@
...
+
+_number_of_particles = SymSymbol("number_of_particles", integer=True)
+
+_amount_of_substance = solve(
+    avogadro_law.law, avogadro_law.mole_count
+)[0].subs(
+    avogadro_law.particles_count, _number_of_particles
+)
+
+_total_mass = _number_of_particles * particle_mass
```
blackyblack, line 35: I wonder if we can derive it from https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/definitions/mass_fraction.py
or from
https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/conservation/mixture_mass_equal_sum_of_components_masses.py

blackyblack (approved):

symplyphysics/laws/chemistry/molar_mass_via_mass_of_single_particle.py
alesanter: Yes, it is possible

blackyblack (approved):

PR #600, opened 2024-04-05
Law: hard spheres potential model

alesanter:


symplyphysics/laws/chemistry/potential_energy_models/hard_spheres_potential.py
```diff
@@ -0,0 +1,48 @@
...
+## r - distance between particles
+## sigma - diameter of sphere
+
+potential = Symbol("potential", units.energy)
+distance = Symbol("distance", units.length)
+sphere_diameter = Symbol("sphere_diameter", units.length)
+
+law = Eq(
+    potential,
+    Piecewise((S.Infinity, distance <= sphere_diameter), (0, distance > sphere_diameter)),
```
blackyblack, line 27: Here we should get infinite energy, not some dimensionless infinity.
Try `S.Infinity * units.energy` or `Quantity(S.Infinity, units.joule)`

symplyphysics/laws/thermodynamics/compressibility_factor_via_intermolecular_force_potential.py
```diff
@@ -49,11 +49,11 @@
 
 # Calculate the compressibility factor for the model of hard spheres
 
-_sphere_radius = Symbol("sphere_radius", units.length, positive=True)
+_sphere_diameter = Symbol("sphere_diameter", units.length, positive=True)
 
 _hard_spheres_potential = Piecewise(
```
blackyblack, line 54: You should take it from the law above.

test/chemistry/potential_energy_models/hard_spheres_potential_test.py
```diff
@@ -0,0 +1,48 @@
...
+    d = Quantity(10 * units.nanometer)
+    return Args(r0=r0, r1=r1, d=d)
+
+
+def test_law_zero_potential(test_args: Args) -> None:
+    result = hard_spheres_potential.calculate_potential(test_args.r0, test_args.d)
+    assert_equal(result, 0)
+
+
+# FIXME: maybe allow infinity too in [here?](https://github.com/blackyblack/symplyphysics/blob/ef7a449974e977a2676e55339d38f4943e7d1ad3/symplyphysics/core/dimensions.py#L107)
```
blackyblack, line 32: I think it raises because you return dimensionless infinity.

symplyphysics/laws/chemistry/potential_energy_models/hard_spheres_potential.py
alesanter: Oh, I see now

alesanter: image
Still doesn't work with infinity

alesanter: `S.Infinity * units.joule` doesn't work either

blackyblack: I've fixed it.

alesanter: Now I see how to manage it, thanks

alesanter: Though this caused a problem in the compressibility factor law since `sympy` now doesn't know that there's infinity in that quantity of ours in this law

alesanter: The problem goes away if I substitute that quantity with proper `sympy.S.Infinity`

blackyblack: Yes, I see that CI is failing. I will look into it.

blackyblack: Done. Take a look.

symplyphysics/laws/thermodynamics/compressibility_factor_via_intermolecular_force_potential.py
```diff
@@ -38,27 +39,32 @@
 intermolecular_force_potential = Function("intermolecular_force_potential", units.energy, real=True)
 temperature = Symbol("temperature", units.temperature, positive=True)
 
+infinite_distance = Quantity(S.Infinity, dimension=units.length)
```
alesanter, line 42: Why do we even treat infinity different from zero, dimension-wise? Both of them consume whatever is multiplied with them: $`0 \cdot x = 0 \, \forall x \in \mathbb{C}`$ and $`\infty \cdot x = \infty \, \forall x \in \bar{\mathbb{C}} \setminus \{0\}`$. So technically infinity would be dimensionless the same way zero is.

symplyphysics/laws/thermodynamics/compressibility_factor_via_intermolecular_force_potential.py
```diff
@@ -38,27 +39,32 @@
...
-_hard_spheres_potential = Piecewise(
-    (S.Infinity, intermolecular_distance < _sphere_radius),
-    (0, intermolecular_distance >= _sphere_radius),
-)
+_hard_spheres_potential = hard_spheres_potential.law.rhs.subs({
+    hard_spheres_potential.distance: intermolecular_distance,
+    hard_spheres_potential.sphere_diameter: _sphere_diameter,
+    # Infinite quantity cannot be integrated properly - convert it to SymPy Infinity
+    # Note, that this way we lose information about its dimension
+    hard_spheres_potential.infinite_potential: hard_spheres_potential.infinite_potential.scale_factor,
```
alesanter, line 62: Plus I don't think we should do extra work taking out infinity from the scale factor every time it appears in expressions. `sympy` will simplify everything on its own if we do not put infinity inside a quantity.

symplyphysics/laws/thermodynamics/compressibility_factor_via_intermolecular_force_potential.py
```diff
@@ -38,27 +39,32 @@
...
 
-_hard_spheres_potential = Piecewise(
-    (S.Infinity, intermolecular_distance < _sphere_radius),
-    (0, intermolecular_distance >= _sphere_radius),
-)
+_hard_spheres_potential = hard_spheres_potential.law.rhs.subs({
+    hard_spheres_potential.distance: intermolecular_distance,
+    hard_spheres_potential.sphere_diameter: _sphere_diameter,
+    # Infinite quantity cannot be integrated properly - convert it to SymPy Infinity
+    # Note, that this way we lose information about its dimension
```
alesanter, line 61: We also will do lose any dimensional information in case of infinite limits. Again, we are already okay with having zero as a limit of integration, yet the information about dimension is lost with it as well.

symplyphysics/laws/thermodynamics/compressibility_factor_via_intermolecular_force_potential.py
blackyblack: I'm not sure about it. 0 joules and 0 meters is the same value, because there is nothing to compare. But infinity is some huge amount of something, so inifinite amount of energy is not the same as infinite distance.
However in most physical applications infinity will be reduced to zero, and its dimension will likely will be erased anyway.

symplyphysics/laws/thermodynamics/compressibility_factor_via_intermolecular_force_potential.py
blackyblack: Zero limit of integration does not lose dimensions. Because it adds up to upper integration limit and zero has a wildcard dimension, ie takes the dimension of its non-zero addend.

symplyphysics/laws/thermodynamics/compressibility_factor_via_intermolecular_force_potential.py
blackyblack: Well, I've made inifinity to be compatible with any dimension, like zero. Not sure if it was a good idea, but code looks simpler, and dimensional analysis with infinity was broken anyway.

alesanter: I think it is good for now. For now it can probably be viewed as a limitation of sympy and our library.

PR #617, opened 2024-04-09
Patch: conversion of dimensionless quantities

alesanter:


symplyphysics/core/convert.py
```diff
@@ -13,3 +13,7 @@ def convert_to(value: SymQuantity, target_unit: SymQuantity) -> Expr:
     assert_equivalent_dimension(value, value.dimension.name, "convert_to",
         target_quantity.dimension)
     return sympify(value.scale_factor) * (1 / sympify(target_quantity.scale_factor))
+
+
+def convert_to_dimensionless(value: SymQuantity) -> float:
```
blackyblack, line 18: `convert_to_float` is a better name.

symplyphysics/definitions/boltzmann_factor.py
```diff
@@ -11,6 +10,7 @@
     symbols,
     clone_symbol,
 )
+from symplyphysics.core.convert import convert_to_dimensionless
```
blackyblack, line 13: Add this import to `init` so it can be imported easily.

blackyblack (approved):

symplyphysics/core/convert.py
alesanter: I agree, sounds better

symplyphysics/core/convert.py
```diff
@@ -13,3 +13,7 @@ def convert_to(value: SymQuantity, target_unit: SymQuantity) -> Expr:
     assert_equivalent_dimension(value, value.dimension.name, "convert_to",
         target_quantity.dimension)
     return sympify(value.scale_factor) * (1 / sympify(target_quantity.scale_factor))
+
+
+def convert_to_float(value: SymQuantity) -> float:
```
blackyblack, line 18: I suggest to also accept `Expr` type of value with automatic conversion to `Quantity`. This way you'll be able to shorten code.

blackyblack (approved):

symplyphysics/core/convert.py
alesanter: Do you mean allowing to write things like `result = law.subs(...); return convert_to_float(result)`, omitting the `Quantity(result)` part in the end? The thing is that `convert_to` accepts only `Quantities` strictly, because it uses the `dimension` field which is not present in `Mul` and other classes.

blackyblack: Yes, but you can see, that we explicitly convert to Quantity almost everywhere. You can do it in your helper function, when you get input parameter as `Expr`

alesanter: I know, I just wanted to make sure I understand you

symplyphysics/core/convert.py
```diff
@@ -1,19 +1,20 @@
 from sympy import Expr, sympify, S
-from sympy.physics.units import Quantity as SymQuantity
+# from sympy.physics.units import Quantity as SymQuantity
 
 from .dimensions import assert_equivalent_dimension
 from .symbols.quantities import Quantity
 
 
-def convert_to(value: SymQuantity, target_unit: SymQuantity) -> Expr:
+def convert_to(value: Expr, target_unit: Expr) -> Expr:
```
blackyblack, line 7: Target unit is under our control. We want to narrow down its type as much as possible. Don't change its type.

symplyphysics/core/convert.py
```diff
@@ -1,19 +1,20 @@
...
     """
     Convert ``value`` to its scale factor with ``value`` unit represented as ``target_unit``.
     """
-    target_quantity = Quantity(target_unit)
-    assert_equivalent_dimension(value, value.dimension.name, "convert_to",
+    value_quantity = value if isinstance(value, Quantity) else Quantity(value)
```
blackyblack, line 11: When you use `isinstance`, its types should be present in function signature. Use `Expr | Quantity` for value type.

symplyphysics/core/convert.py
alesanter: Okay

symplyphysics/core/convert.py
alesanter: But `Quantity | Expr` is virtually equivalent to `Expr` via MRO since `Quantity` is a subclass of `Expr`

blackyblack: Agree.

symplyphysics/core/convert.py
blackyblack: Hm, good point. Maybe it can be safely converted to SymQuantity?

alesanter: Not sure why. Tried to understand the cause but it seems like that's some quirk of `mypy`. But this is a reason to leave `Expr` to be the type of `target_unit`.

blackyblack: Ok, let's keep it this way.

blackyblack (approved):

PR #628, opened 2024-04-11
Derive: free energy differential

alesanter:


symplyphysics/laws/thermodynamics/free_energy_differential.py
```diff
@@ -46,7 +51,57 @@
     -1 * entropy * temperature_change - pressure * volume_change + chemical_potential * particle_count_change,
 )
 
-# TODO: derive from the definition of Helmholtz free energy and internal energy differential
+# Derive from the definition of Helmholtz free energy and internal energy differential
```
alesanter, line 54: I suppose we need some better machinery to work with differentials, but for now, I guess, this suffices.

symplyphysics/laws/thermodynamics/free_energy_differential.py
```diff
@@ -46,7 +51,57 @@
...
+    free_energy_def.internal_energy: _internal_energy,
+    free_energy_def.temperature: _temperature,
+    free_energy_def.entropy: entropy,
+})
+
+_free_energy_change = (
+    _free_energy.diff(_temperature) * temperature_change
+    + _free_energy.diff(entropy) * _entropy_change
+    + _free_energy.diff(_volume) * volume_change
+    + _free_energy.diff(_particle_count) * particle_count_change
```
blackyblack, line 74: What is this formula?

symplyphysics/laws/thermodynamics/free_energy_differential.py
```diff
@@ -46,7 +51,57 @@
...
+_internal_energy_diff_volume = _internal_energy_diff.subs({
+    _entropy_change: 0, volume_change: 1, particle_count_change: 0
+})
+
+_internal_energy_diff_particle_count = _internal_energy_diff.subs({
+    _entropy_change: 0, volume_change: 0, particle_count_change: 1
+})
+
+_free_energy_change = _free_energy_change.subs({
+    _internal_energy.diff(entropy): _internal_energy_diff_entropy,
```
blackyblack, line 99: I think it is better to use `Derivative(_internal_energy(entropy), entropy)` instead of `diff()` if you are not really interested in the result.

symplyphysics/laws/thermodynamics/free_energy_differential.py
alesanter: Here I'm looking to find the differential of the free energy. I don't think that `sympy` can do it, so I have to do it by hand

symplyphysics/laws/thermodynamics/free_energy_differential.py
alesanter: `_internal_energy` is a symbol so there is no differentiation happening anyway but it looks less cluttered

symplyphysics/laws/thermodynamics/free_energy_differential.py
blackyblack: Add some hint about it.

blackyblack (approved):

PR #629, opened 2024-04-11
Derive: dimensionless van der Waals equation

alesanter:


symplyphysics/laws/thermodynamics/equations_of_state/van_der_waals/dimensionless_equation.py
```diff
@@ -30,7 +33,33 @@
...
+_critical_molar_volume_expr = critical_molar_volume.law.rhs.subs({
+    critical_molar_volume.molecules_volume_parameter: equation.molecules_volume_parameter,
+})
+
+_critical_temperature_expr = critical_temperature.law.rhs.subs({
+    critical_temperature.bonding_forces_parameter: equation.bonding_forces_parameter,
+    critical_temperature.molecules_volume_parameter: equation.molecules_volume_parameter,
+})
+
+_reduced_eqn = equation.law.subs({
```
alesanter, line 52: I tried to pull it off with equations for reduced quantities and then plugging it into `solve` so that `sympy` performed the necessary substitutions, but nothing seemed to trigger them

symplyphysics/laws/thermodynamics/equations_of_state/van_der_waals/dimensionless_equation.py
```diff
@@ -30,7 +33,33 @@
...
+    equation.pressure: reduced_pressure * _critical_pressure_expr,
```
blackyblack, line 53: Ideally we should use a law to derive this. At least, point user that `p* = p/p_c` from the description above.

symplyphysics/laws/thermodynamics/equations_of_state/van_der_waals/dimensionless_equation.py
```diff
@@ -30,7 +33,33 @@
...
+    equation.volume: reduced_volume * _critical_molar_volume_expr * equation.amount_of_substance,
```
blackyblack, line 54: It seems, we have molar heat capacity definition, but no other `molar` parameters. Can we write some general law of `molar` something and use it here for calculating critical volume?

alesanter: You mean like a general law that relates a non-molar and a molar quantity, say, $`X = X_m \cdot n`$, with $`X`$ and $`X_m`$ being `sympy.Symbol`s?

blackyblack: Yeah

symplyphysics/laws/thermodynamics/equations_of_state/van_der_waals/dimensionless_equation.py
blackyblack: I see there is some common pattern emerging. We have `molar` something, when we divide by amount of moles. We have `volumetric_density` something, when we divide by volume. We have `specific` something, when we divide by mass. And here we have `reduced` something, when we divide by critical value.

alesanter: Agree with all but `reduced`. It doesn't need to be a _critical_ value but can be any quantity of the same dimension as the original one that makes use of some characteristic parameters. E.g. [Lennard-Jones potential](https://en.wikipedia.org/wiki/Lennard-Jones_potential#Dimensionless_(reduced_units)).

alesanter: So with reduced values there will be separate definitions for each model that uses them. Like here, it's probably better to have three separate laws. But the `molar` and `specific` ones are all good.

blackyblack: Well, it's up to you what you think looks better.

blackyblack (approved):

PR #635, opened 2024-04-11
transmission_matrix_of_t_type_circuit

Zulusssss:
Add law and tests.

### DIFF
--- a/symplyphysics/laws/electricity/transmission_lines/transmission_matrix_of_t_type_circuit.py
+++ b/symplyphysics/laws/electricity/transmission_lines/transmission_matrix_of_t_type_circuit.py
@@ -1,0 +1,101 @@
+from sympy import Eq, solve, Matrix
+from symplyphysics import (
+    units,
+    Quantity,
+    Symbol,
+    print_expression,
+    validate_input,
+    dimensionless,
+    convert_to_float
+)
+from symplyphysics.core.dimensions import assert_equivalent_dimension
+from symplyphysics.core.expr_comparisons import expr_equals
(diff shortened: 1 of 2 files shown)


symplyphysics/laws/electricity/transmission_lines/transmission_matrix_of_t_type_circuit.py
```diff
@@ -0,0 +1,101 @@
...
+)
+from symplyphysics.core.dimensions import assert_equivalent_dimension
+from symplyphysics.core.expr_comparisons import expr_equals
+from symplyphysics.laws.electricity.transmission_lines import transmission_matrix_for_a_series_load_in_line as series_law
+from symplyphysics.laws.electricity.transmission_lines import transmission_matrix_for_a_parallel_load_in_line as parallel_law
+
+## Description
+## The transmission parameters matrix is one of the ways to describe a microwave device. The ABCD-parameters of the device act as elements
+## of this matrix. The matrix equation relates the input voltage and input current to the output voltage and output current.
+## The T-type circuit consists of the first impedance connected in series, the third impedance connected in parallel, and the
```
blackyblack, line 19: Shouldn't second impedance be parallel? Why does third impedance come before second?

Zulusssss: > Shouldn't second impedance be parallel? Why does third impedance come before second?

It's just that such designations were on the diagram for which the formula was obtained. The impedances that are set the same way are called the first and second, and the remaining impedance is called the third.

PR #638, opened 2024-04-13
Core: helper function to construct linear function from two points

alesanter:


symplyphysics/core/geometry/line.py
```diff
@@ -0,0 +1,20 @@
+from sympy import symbols, Point, Line, Expr, solve
+
+def two_point_function(p1: Point, p2: Point, x: Expr) -> Expr:
```
blackyblack, line 3: You can put points in tuple.

blackyblack, line 3: Can we narrow down `x` type to `Symbol`?

symplyphysics/core/geometry/line.py
```diff
@@ -0,0 +1,20 @@
...
+    """Constructs a linear function of ``x`` using two points ``p1`` and ``p2``.
+    
+    Returns ``NotImplemented`` if the line equation does not depend on ``y``.
+    """
+
+    xsym = x
```
blackyblack, line 9: Don't save to temporary variable.

symplyphysics/core/geometry/line.py
```diff
@@ -0,0 +1,20 @@
...
+
+    x, y = symbols("x y")
```
blackyblack, line 11: Avoid `x` name to not shadow the user variable.

symplyphysics/core/geometry/line.py
```diff
@@ -0,0 +1,20 @@
...
+    line = Line(p1, p2)
```
blackyblack, line 12: Use `Line2D` instead.

symplyphysics/core/geometry/line.py
```diff
@@ -0,0 +1,20 @@
...
+    equation = line.equation(x=x, y=y)  # type: ignore[attr-defined]
```
blackyblack, line 13: Remove linter hint.

symplyphysics/core/geometry/line.py
```diff
@@ -0,0 +1,20 @@
...
+    
+    y_functions = solve(equation, y)
+
+    if not y_functions:
+        return NotImplemented
```
blackyblack, line 18: Throw exception.

symplyphysics/core/geometry/line.py
```diff
@@ -0,0 +1,20 @@
...
+
```
blackyblack, line 19: Also raise on the same Point instead of 2 different points.

test/core/geometry/line_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+def test_nonzero_intercept() -> None:
+    p1 = Point(1, 1)
+    p2 = Point(2, 3)
+    x = Symbol("x")
+    y = two_point_function(p1, p2, x)
+    assert expr_equals(y, 2 * x - 1)
+
+
+def test_parallel_to_y_axis() -> None:
```
blackyblack, line 21: Add test for same value in both points and for parallel to x axis.

symplyphysics/core/geometry/line.py
alesanter: It seemed reasonable to me to have `x` as a parameter of this function. Maybe `x_sym` can be fine too.

blackyblack: You can keep it `x`. But rename temporary `x, y` symbols.

symplyphysics/core/geometry/line.py
alesanter: That would be too narrow a type. We might want to have, say, `f(x)` as the `x`, and the closest type it shares with `Symbol` in the MRO hierarchy is `Expr`.

alesanter: You mean to have `tuple[Expr, Expr]` as the type of `p1` and `p2` or have a union of that and `Point2D`?

symplyphysics/core/geometry/line.py
alesanter: `Line2D` would have raised it already, so no need.

symplyphysics/core/geometry/line.py
blackyblack: `tuple[Point, Point]`

blackyblack: I doubt we'll ever want `f(x)` as parameter here.

symplyphysics/core/geometry/line.py
blackyblack: So we need a test to show it.

symplyphysics/core/geometry/line.py
alesanter: How is that better than having two parameters?

alesanter: Okay, we can have `Symbol` for now

blackyblack: You won't have point1 and point2. Just some set of points.

alesanter: I think it's better to restrict it to two parameters. It would make sense to have a tuple if we were to allow any arbitrary collection of points of any length. But we specifically construct the equation using two points. If we have a tuple of points on hand, say `points`, we can always pass `*points` as the argument.

blackyblack: Tuple is not of arbitrary length. Tuple[Point, Point] is for 2 points. Usually having variables with names like `var1`, `var2` is a red flag. But it is not a big deal so choose yourself if you want to pack them in a tuple or not.

alesanter: All of `sympy` uses `p1` and `p2` arguments, at least up to the point I've seen, so I wanted it to be similar to that.

blackyblack: Ok

alesanter: It's done then

blackyblack: But other issues are not fixed.

alesanter: Ah, forgot to sync

symplyphysics/core/geometry/line.py
```diff
@@ -0,0 +1,18 @@
+from sympy import symbols, Point2D, Line2D, Expr, solve
+
+def two_point_function(p1: Point2D | tuple[Expr, Expr], p2: Point2D | tuple[Expr, Expr], x: Expr) -> Expr:
```
blackyblack, line 3: I don't think having point as tuple adds any value. Remove this type.

blackyblack (approved):

PR #639, opened 2024-04-13
Law: extensive quantity via corresponding molar quantity

alesanter:


symplyphysics/laws/quantities/quantity_is_molar_quantity_times_amount_of_substance.py
```diff
@@ -0,0 +1,47 @@
...
+    molar_quantity_: Quantity,
+    amount_of_substance_: Quantity,
+) -> Quantity:
+    result_expr = law.rhs.subs({
+        molar_quantity: molar_quantity_,
+        amount_of_substance: amount_of_substance_,
+    })
+    result = Quantity(result_expr)
+
+    assert dimsys_SI.equivalent_dims(
```
blackyblack, line 42: Use `assert_equivalent_dimension`

blackyblack (approved):

symplyphysics/laws/quantities/quantity_is_molar_quantity_times_amount_of_substance.py
alesanter: I actually now think there is no reason to check the dimensions because it is impossible to return the wrong dimension.

blackyblack: Yes, I thought about it. But still might be nice to indicate the resulting dimension.

alesanter: Okay, added it

symplyphysics/laws/quantities/quantity_is_molar_quantity_times_amount_of_substance.py
```diff
@@ -0,0 +1,49 @@
...
+    amount_of_substance_: Quantity,
+) -> Quantity:
+    result_expr = law.rhs.subs({
+        molar_quantity: molar_quantity_,
+        amount_of_substance: amount_of_substance_,
+    })
+    result = Quantity(result_expr)
+
+    assert_equivalent_dimension(
+        molar_quantity_,
```
blackyblack, line 43: Check result.

symplyphysics/laws/quantities/quantity_is_molar_quantity_times_amount_of_substance.py
```diff
@@ -0,0 +1,49 @@
...
+        "molar_quantity_",
```
blackyblack, line 44: We usually call it `return`

symplyphysics/laws/quantities/quantity_is_molar_quantity_times_amount_of_substance.py
```diff
@@ -0,0 +1,49 @@
...
+        "calculate_extensive_quantity",
+        result.dimension / units.amount_of_substance,
```
blackyblack, line 46: It doesn't make sense. Check the resulting dimension.

symplyphysics/laws/quantities/quantity_is_molar_quantity_times_amount_of_substance.py
alesanter: image
The assertion function raises error messages about arguments, not returns. That's why I reformulated it in terms of argument, I didn't want to change `assert_equivalent_dimension`. If we change the error texts to be about the return value, it will be just alright.

blackyblack: You can see that we use it for returned values as well: https://github.com/blackyblack/symplyphysics/blob/1d23238a7aa98f7410bd9e34b1d1e2f6ae2556d1/symplyphysics/core/quantity_decorator.py#L83
I agree that text is a bit misleading but it's ok.

alesanter: Hm, okay, now I see

blackyblack (approved):

PR #647, opened 2024-04-15
Law: chemical potential is internal energy derivative

alesanter:


symplyphysics/laws/thermodynamics/chemical_potential_is_particle_count_derivative_of_internal_energy.py
```diff
@@ -0,0 +1,71 @@
...
+    validate_output,
+)
+from symplyphysics.core.geometry.line import two_point_function
+
+# Description
+## The chemical potential of a system is the amount of energy the system absorbs or releases
+## due to the introduction of a particle into the system, i.e. when the particle count increases
+## by one.
+
+# Law: mu = (dU/dN)_(S, V)
```
blackyblack, line 19: Now I think if it was a good idea to separate laws for different kind of energies.

alesanter: In my opinion it's better to have 4 easily manageable specific laws than 1 general law.

PR #653, opened 2024-04-15
Law: Wien distribution of blackbody radiation

alesanter:


symplyphysics/laws/waves/blackbody_radiation/wien/spectral_energy_density.py
```diff
@@ -0,0 +1,60 @@
+from sympy import Eq, exp, pi
```
blackyblack, line 1: It seems that you prefer to separate equations by marketing names, eg `planck` or `wien`. But how is user supposed to find proper equation for his needs? Can we somehow describe, how they are different from each other? And we can use this difference as folder name.

blackyblack (approved):

symplyphysics/laws/waves/blackbody_radiation/wien/spectral_energy_density.py
alesanter: I'm not sure how to adequately name the folders so that they convey the "law works correctly for high/low/any frequencies" meaning

alesanter: We can also get rid of folders entirely and just mention the distinction in file names.

blackyblack: Good idea, if we expect just one related law.

PR #654, opened 2024-04-15
Refactor: use two_point_function in calculations

alesanter:


symplyphysics/definitions/thermodynamic_compressibility.py
```diff
@@ -55,9 +56,10 @@ def calculate_compressibility(
 ) -> Quantity:
     # The value of the volume is calculated in the `pressure_after_` point
 
-    volume_function = (
-        volume_before_ +
-        (volume_after_ - volume_before_) * (pressure - pressure_before_) / (pressure_after_ - pressure_before_)
+    volume_function = two_point_function(
+        Point2D(pressure_before_, volume_before_),
+        Point2D(pressure_after_, volume_after_),
+        x=pressure,
```
blackyblack, line 62: AFAIR `x` is not position independent parameter. You can simply put `pressure` here.

alesanter: Well yes, but for me this is a reminder of what that positional parameter is about

blackyblack (approved):

PR #662, opened 2024-04-17
Define & derive: coefficients of thermal expansion

alesanter:


symplyphysics/definitions/linear_coefficient_of_thermal_expansion.py
```diff
@@ -0,0 +1,63 @@
...
+# Conditions
+## - Pressure must be constant during the expansion process
+
+linear_expansion_coefficient = Symbol("linear_expansion_coefficient", 1 / units.temperature)
+length = Function("length", units.length)
+temperature = symbols.thermodynamics.temperature
+
+definition = Eq(
+    linear_expansion_coefficient,
+    Derivative(length(temperature), temperature) / length(temperature),
```
blackyblack, line 33: I am quite sure that `temperature` as derivative parameter, and `temperature` in `length(temperature)` are different temperatures.
For example, most sources use `L0` instead of `length(temperature)`, which is basically `length(temperature_before)`

blackyblack: We have similar laws, with the same issue.

symplyphysics/laws/thermodynamics/volumetric_and_linear_expansion_coefficients_in_isotropic_materials.py
```diff
@@ -21,7 +26,39 @@
...
+_temperature_change = symbols("temperature_change")
+
+_final_length_expr = _initial_length + _length_change
+_initial_volume_expr = _initial_length**3
+_final_volume_expr = _final_length_expr**3
+
+_volumetric_expr = volumetric_def.definition.rhs.subs({
+    volumetric_def.volume(volumetric_def.temperature).diff(volumetric_def.temperature): _volume_change / _temperature_change,
+    _volume_change: _final_volume_expr - _initial_volume_expr,
+    volumetric_def.volume(volumetric_def.temperature): _initial_volume_expr,
```
blackyblack, line 42: Not justified. It assumes `V(t)` is `V(t0)` which I suggested above.

symplyphysics/laws/thermodynamics/volumetric_and_linear_expansion_coefficients_in_isotropic_materials.py
```diff
@@ -21,7 +26,39 @@
...
+})
+
+# Note: if dimension changes are not small, the third and fourth terms of the following series expansion
+# can be taken into account.
+_volumetric_expr_series = _volumetric_expr.series(_length_change, 0, 2).removeO()
+
+_linear_eqn = linear_def.definition.subs({
+    linear_def.length(linear_def.temperature).diff(linear_def.temperature): _length_change / _temperature_change,
+    linear_def.length(linear_def.temperature): _initial_length,
```
blackyblack, line 51: Same here.

symplyphysics/definitions/linear_coefficient_of_thermal_expansion.py
alesanter: Not sure if I follow you. Both the derivative $`\frac{d L}{d T}`$ and the original function $`L(T)`$ are calculated at the same temperature point, in theory. The practical (engineering or laboratory) approximation of the derivative might differ in their choice of points and such, but this is a formal mathematical definition.

blackyblack: If you use L(t) like described, you will get expansion_function(t). I'm not sure it is really useful this way. Most people expect coefficient at some temperature point, ie at L(t0).
But we cannot take derivative from L(t0), so L(t) in derivative and L(t) after it are two different entities.

alesanter: If this law were about finite changes in temperature and size, i.e. being something like `(length_change / length) / temperature_change`, then yes, we would have to specify the exact temperature points we are calculating these values at. But here the changes are infinitesimal, the functions are continuous, and we can and do actually calculate the value of the derivative of size at the same temperature as the size function. It's basically the rate of change of size divided by the size, both at the same temperature. For engineering or practical purposes, this "true" derivative will be replaced with a discrete one, which is the finite case I wrote above.

alesanter: I understand that the notation $`\frac{d L}{d T}`$ is confusing, but this is basically the same as $`\left.\frac{d L(T')}{d T'}\right|_{T' = T}`$. I've tried this in `sympy` just now and it too makes no difference between the two notations.

symplyphysics/laws/thermodynamics/volumetric_and_linear_expansion_coefficients_in_isotropic_materials.py
alesanter: I'll get back to you on this

symplyphysics/definitions/linear_coefficient_of_thermal_expansion.py
blackyblack: I'll think about it. I tend to agree with you.
Maybe calling the resulting value a `function` would resolve my confusion.

symplyphysics/laws/thermodynamics/volumetric_and_linear_expansion_coefficients_in_isotropic_materials.py
alesanter: And about this part here, even if we choose `_final_volume_expr` to be the denominator, it is the same as `_initial_volume_expr` in the `_volume_change -> 0` limit. Or, $`\frac{1}{V + \Delta V} = \frac{1}{V} \frac{1}{1 + \frac{\Delta V}{V}} = \frac{1}{V} \left(1 + \sum_{i \ge 1} \left( - \frac{\Delta V}{V} \right)^i \right) = \frac{1}{V}`$ in the limit $`\Delta V \to 0`$.

symplyphysics/definitions/linear_coefficient_of_thermal_expansion.py
alesanter: We already defined `length` to be a `Function` and also always write `length(temperature)`. To me it's already a good indication that it's a function.

blackyblack: I meant `linear_expansion_coefficient`

blackyblack: I meant its name :)

alesanter: I think that would lead to confusion. It's clear it's not a constant, is it

blackyblack: Not really. Coefficient looks like some constant value.

symplyphysics/laws/thermodynamics/volumetric_and_linear_expansion_coefficients_in_isotropic_materials.py
blackyblack: Still we should somehow justify how we choose this parameter. You can either add some explanation, or add condition that `volumetric_def.volume(temperature) == linear_def.length(temperature)**3` and derive from it.

symplyphysics/definitions/linear_coefficient_of_thermal_expansion.py
alesanter: I suggest writing it in the description. Also, for it to be a constant value, the length function must be an exponential function, which is rarely the case.

symplyphysics/laws/thermodynamics/volumetric_and_linear_expansion_coefficients_in_isotropic_materials.py
alesanter: Sure

symplyphysics/definitions/linear_coefficient_of_thermal_expansion.py
blackyblack: For constant value it should be taken at some given point t0. That's why I thought t is different under and out of derivative.

alesanter: Why does having $`L(t_0)`$ in the denominator make this coefficient constant? The derivative $`\frac{dL}{dT}`$ is still a function of temperature in general. It _is_ outside the derivative, but they both depend on the same temperature "symbol".

Also, we seem to cute different sources. As far as I can see, Russian ones (I looked in Sivuhin's books) use $`L(0 \degree \mathrm{C})`$ while English ones (say [Springer](https://link.springer.com/chapter/10.1007/978-3-662-02954-1_22)).

PR #664, opened 2024-04-17
Law: work integral in thermodynamics

alesanter:


symplyphysics/laws/thermodynamics/work_is_volume_integral_of_pressure.py
```diff
@@ -0,0 +1,81 @@
...
+# Prepare law for integration
+_infinitesimal_law = work_law.law.subs({
+    work_law.infinitesimal_work_done: 1,
+    work_law.infinitesimal_volume_change: 1,
+    work_law.pressure_inside_system: pressure(volume),
+})
+
+_integrated_law = Eq(
+    _infinitesimal_law.lhs.integrate(work),
+    _infinitesimal_law.rhs.integrate((volume, volume_before, volume_after))
```
blackyblack, line 46: I think that both sides of equation should have the same integration limits.

symplyphysics/laws/thermodynamics/work_is_volume_integral_of_pressure.py
```diff
@@ -34,19 +34,26 @@
...
 
 _integrated_law = Eq(
-    _infinitesimal_law.lhs.integrate(work),
+    _infinitesimal_law.lhs.integrate((volume, volume_before, volume_after)),
     _infinitesimal_law.rhs.integrate((volume, volume_before, volume_after))
 )
 
-_work_derived = solve(_integrated_law, work)[0]
+_work_derived = solve(
+    (Eq(work, _work(volume_after) - _work(volume_before)), _integrated_law),
```
blackyblack, line 53: Can we somehow justify that work equals to `work(volume_after) - _work(volume_before)` ?

alesanter: The fact that we have the `_work` function at different volume points is mostly an artefact of integration. `_work` is nothing but an antiderivative which doesn't carry much meaning, except its difference between two volume points is exactly the work done by the system as its volume changes between those two points. We can just as well set `_work(volume_before)` to `0`, and `_work(volume_after)` to `work` and it will be correct too.

blackyblack: I think it is a good idea to subs work_before with 0, because it's quite obvious. We'll have just one work as a result and we can more or less safely substitute it for `work`.
I think there is some more formal condition. Something about trajectory independence of work.

alesanter: I don't think this is a required condition. If we go from infinitesimal changes in volume and infinitesimal amounts work and then sum them up, on the left side we get $`\sum_i \delta W_i = W`$ and the right side evaluates to an integral $`\int_{V_i}^{V_f} p(V) dV`$. So far, no need of any work antiderivatives or quasi-static conditions. That's the reason for what I originally wrote there.

alesanter: I think so

symplyphysics/laws/thermodynamics/work_is_volume_integral_of_pressure.py
blackyblack, line 46: I'm still not satisfied with this proof. For example we can change RHS integration limits, eg `(volume, volume_before, volume_after / 2)` yet the total work won't change. It means that left and right hand side integrations are not linked to each other.

alesanter: Then the best option would be to have `_work(volume_before) = 0, _work(volume_after) = work`

PR #666, opened 2024-04-17
Derive: chemical potential is Gibbs energy per particle

alesanter:


symplyphysics/laws/thermodynamics/chemical_potential_is_gibbs_energy_per_particle.py
```diff
@@ -24,7 +28,44 @@
...
+_temperature = SymSymbol("temperature")
+_factor = SymSymbol("factor")
+
+_gibbs_energy_eqn = Eq(gibbs_energy, _gibbs_energy(_temperature, _pressure, particle_count))
+
+# Particle count and Gibbs energy are extensive quantities, whereas temperature and pressure
+# are intensive, which means that the latter do not scale with particle count, unlike Gibbs energy.
+_gibbs_energy_factored_eqn = _gibbs_energy_eqn.subs({
+    particle_count: _factor * particle_count,
+    gibbs_energy: _factor * gibbs_energy,
```
blackyblack, line 44: Maybe it's better to write Gibbs energy for 1 particle?

alesanter: You mean the `gibbs_energy` symbol? I don't really understand what you mean

alesanter: But we need to show that the Gibbs energy scales alongside with particle count. It doesn't follow from anything.

blackyblack: And you didn't justify it either.

blackyblack: Cool idea. But we should make sure it is really so.

alesanter: You mean the fact that thermodynamic potentials (Gibbs energy in particular) are homogeneous functions is not evident? That law I proposed is this property exactly.

blackyblack: Yes but we should rely on some physical reality in the law. We cannot simply invent them. That's why I suggested to look for some evidence that such law exists.

alesanter: Well I'm not pulling it out of thin air either. All thermodynamic potentials scale with particle number, that's their property. You can see [Euler relations](https://en.wikipedia.org/wiki/Thermodynamic_potential#Euler_relations) for more.

blackyblack: I see that other potentials scale with particle number but it is not a direct scale. For example, doubling particle number does not double inner energy. But for Gibbs it seems, there is direct relation.

alesanter: Well, internal energy requires scaling of all of its natural variables because they are all extensive. The only extensive natural variable of the Gibbs potential is the particle number. In that law I proposed I only wrote about Gibbs energy. All other potentials would require separate laws.

alesanter: We actually won't need this law since there's an equivalent Euler relation for Gibbs energy

alesanter: Although it might be useful to implement the scaling laws

PR #673, opened 2024-04-18
Derive: adiabatic equation

alesanter:


symplyphysics/laws/thermodynamics/zero_heat_transfer.py
```diff
@@ -25,26 +41,119 @@
...
+_temperature_change = SymSymbol("temperature_change")
+_pressure_change = SymSymbol("pressure_change")
+_volume_change = SymSymbol("volume_change")
+
+_internal_energy_expr = internal_energy_law.law.rhs.subs({
+    internal_energy_law.temperature: _temperature,
+    internal_energy_law.isochoric_heat_capacity: _isochoric_heat_capacity,
+})
+
+_internal_energy_change_expr = _internal_energy_expr.diff(_temperature) * _temperature_change
```
blackyblack, line 77: Here it assumes that heat capacity is constant. But generally it is a function of temperature.

symplyphysics/laws/thermodynamics/zero_heat_transfer.py
```diff
@@ -25,26 +41,119 @@
...
+    _pressure: _parameterized_pressure(t),
+    _volume: _parameterized_volume(t),
+}).diff(t).subs({
+    _parameterized_pressure(t).diff(t): _pressure_change,
+    _parameterized_volume(t).diff(t): _volume_change,
+    _parameterized_pressure(t): _pressure,
+    _parameterized_volume(t): _volume,
+})
+
+_mayers_relation = mayers_relation.law.subs(
```
blackyblack, line 107: There is typo in Mayers relation description. `at constant temperature` instead of `at constant volume`

symplyphysics/laws/thermodynamics/zero_heat_transfer.py
```diff
@@ -25,26 +41,119 @@
...
+
+_first_law_eqn = first_law.law.subs({
+    first_law.internal_energy_change: _internal_energy_change_expr,
+    first_law.heat_supplied_to_system: 0,
+    first_law.work_done_by_system: _work_change_expr,
+})
+
+# Expressing the temperature differential through pressure and volume differentials
+
+_parameterized_pressure = SymFunction("pressure")
```
blackyblack, line 92: Use SymPy `symbols(cls=SymFunction)` because Pylint complains otherwise.

symplyphysics/laws/thermodynamics/zero_heat_transfer.py
alesanter: True. I should refactor that law so that it has differentials instead of integrated quantities and this proof wouldn't need an unnecessary conditon. Originally it is $`\mathrm{d}U = C_V \mathrm{d}T`$.

alesanter: Opened #679

symplyphysics/laws/thermodynamics/zero_heat_transfer.py
alesanter: Sure. I just usually try to avoid importing `sympy.symbols` because of the name clash with `symplyphysics.symbols` (which is kind of annoying).

symplyphysics/laws/thermodynamics/zero_heat_transfer.py
blackyblack: Meged

blackyblack (approved):

PR #688, opened 2024-04-22
finding_filter_order_from_equation

Zulusssss:
Add law and tests

### DIFF
--- a/symplyphysics/laws/electricity/circuits/filter_order_from_distortion_and_frequencies.py
+++ b/symplyphysics/laws/electricity/circuits/filter_order_from_distortion_and_frequencies.py
@@ -1,0 +1,62 @@
+from math import ceil
+from sympy import (
+    Expr,
+    Eq,
+    solve,
+)
+from symplyphysics import (
+    units,
+    Quantity,
+    Symbol,
+    Function,
+    print_expression,
+    validate_input,
+    validate_output,
+    dimensionless,
+    convert_to_float
+)
+
+# Description
+## The approximation of the power transmission coefficient of a normalized low-pass filter is given by approximating functions of the order of n.
+## Bandwidth distortion determines the maximum distortion in the bandwidth. In other words, to determine the level of ripples in the bandwidth.
+## The band-stop distortion distortion determines the level of distortion in the band-stop.
+
+# Law: F(fs, n) = es / ep
+## Where:
+## F(f, n) - approximating function of the order of n,
+## es - band-stop distortion,
+## ep - bandwidth distortion,
+## fs - frequency of the band-stop.
+
+band_stop_frequency = Symbol("band_stop_frequency", units.frequency)
+filter_order = Symbol("filter_order", dimensionless)
+filter_function = Function("filter_function", dimensionless)
+bandwidth_distortion = Symbol("bandwidth_distortion", dimensionless)
+band_stop_distortion = Symbol("band_stop_distortion", dimensionless)
+
+law = Eq(filter_function(band_stop_frequency, filter_order), band_stop_distortion / bandwidth_distortion)
+
+
+def print_law() -> str:
+    return print_expression(law)
+
+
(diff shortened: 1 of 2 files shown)


symplyphysics/laws/electricity/circuits/filter_order_from_distortion_and_frequencies.py
```diff
@@ -0,0 +1,62 @@
...
+    validate_input,
+    validate_output,
+    dimensionless,
+    convert_to_float
+)
+
+# Description
+## The approximation of the power transmission coefficient of a normalized low-pass filter is given by approximating functions of the order of n.
+## Bandwidth distortion determines the maximum distortion in the bandwidth. In other words, to determine the level of ripples in the bandwidth.
+## The band-stop distortion distortion determines the level of distortion in the band-stop.
```
blackyblack, line 22: Duplicate `distortion` and bad explanation.

symplyphysics/laws/electricity/circuits/filter_order_from_distortion_and_frequencies.py
```diff
@@ -0,0 +1,62 @@
...
+def apply_filter_function(filter_function_: Expr) -> Expr:
+    applied_law = law.subs(filter_function(band_stop_frequency, filter_order), filter_function_)
+    return applied_law
+
+
+@validate_input(band_stop_frequency_=band_stop_frequency,
+                bandwidth_distortion_=bandwidth_distortion,
+                band_stop_distortion_=band_stop_distortion)
+@validate_output(filter_order)
+def calculate_order(filter_function_: Expr, band_stop_frequency_: Quantity, bandwidth_distortion_: float,
```
blackyblack, line 53: I think it is a bad idea to calculate parameter of a function, using this law. You already know this parameter, when you build a function. I suggest to rename this law to `filter_function_from_distortion_and_frequencies` and calculate here either `F(f, n)` or some of distortions.

Zulusssss: > I think it is a bad idea to calculate parameter of a function, using this law. You already know this parameter, when you build a function. I suggest to rename this law to `filter_function_from_distortion_and_frequencies` and calculate here either `F(f, n)` or some of distortions.

Usually, for a filter, we set the distortion in the bandwidth and delay, as well as the cutoff frequency and the suppression frequency. Next, we select one of the approximating polynomials of order "n". But we don't know exactly what order to choose in order for the filter to meet our requirements (distortion and frequencies). Therefore, we substitute our polynomial into this equation and find the minimum suitable order.

The problem of finding a specific polynomial F(w, n) with a given order is not uniquely solvable, since there are many such polynomials.

It is possible to find distortions by knowing the frequencies and the polynomial of a specific order, but this is strange. Do that?

test/electricity/circuits/filter_order_from_distortion_and_frequencies_test.py
```diff
@@ -0,0 +1,53 @@
...
+
+    return Args(filter_function=filter_function,
+        band_stop_frequency=band_stop_frequency,
+        bandwidth_distortion=bandwidth_distortion,
+        band_stop_distortion=band_stop_distortion)
+
+
+def test_basic_order(test_args: Args) -> None:
+    result = order_law.calculate_order(test_args.filter_function, test_args.band_stop_frequency, test_args.bandwidth_distortion, test_args.band_stop_distortion)
+    assert_equal(result, 3, tolerance=0.01)
```
blackyblack, line 37: I think you do not need tolerance, because it is exactly 3.

symplyphysics/laws/electricity/circuits/filter_order_from_distortion_and_frequencies.py
blackyblack: On a second thought, your current solution is good.

blackyblack (approved):

PR #699, opened 2024-04-24
Law: entropy derivative via volume derivative

alesanter:


symplyphysics/laws/thermodynamics/entropy_derivative_via_volume_derivative.py
```diff
@@ -0,0 +1,99 @@
...
+_gibbs_energy_expr = sym_symbols("gibbs_energy", cls=SymFunction)(temperature, pressure)
+
+_gibbs_energy_diff_temperature = gibbs_energy_differential.law.rhs.subs({
+    gibbs_energy_differential.temperature_change: 1,
+    gibbs_energy_differential.pressure_change: 0,
+    gibbs_energy_differential.particle_count_change: 0,
+    gibbs_energy_differential.entropy: entropy(temperature, pressure)
+})
+
+_gibbs_energy_diff_temperature_diff_pressure = _gibbs_energy_diff_temperature.diff(pressure)
```
blackyblack, line 58: Could you explain, what you're doing here (+ code above)?

alesanter: Will do

symplyphysics/laws/thermodynamics/entropy_derivative_via_volume_derivative.py
```diff
@@ -0,0 +1,105 @@
...
+
+# Since Gibbs energy is a differentiable function of all its parameters, the can apply
+# the symmetry of second derivatives, which is that for any sufficiently differentiable
+# function one can exchange the order of taking partial derivatives: d(df/dx)/dy = d(df/dy)/dx.
+# The expression for the second derivative of Gibbs energy must be the same whether one first
+# takes it w.r.t. temperature and then w.r.t. pressure or the other way round.
+
+_gibbs_energy_diff_temperature = gibbs_energy_differential.law.rhs.subs({
+    gibbs_energy_differential.temperature_change: 1,
+    gibbs_energy_differential.pressure_change: 0,
```
blackyblack, line 59: Thanks for the description above. But is it safe to do so? You set `pressure_change` to 0 and in the code below you differentiate over pressure.

alesanter: It's safe. The differential $`dp`$ applies only to the calculation of the first-order differential $`dG`$. It doesn't imply that the pressure stays constant in the whole expression. The factor at $`dp`$, which is basically $`\left(\frac{dG}{dp}\right)_T`$, can itself depend on pressure.

symplyphysics/laws/thermodynamics/entropy_derivative_via_volume_derivative.py
```diff
@@ -0,0 +1,105 @@
...
+law = Eq(
+    Derivative(entropy(temperature, pressure), pressure),
+    -1 * Derivative(volume(temperature, pressure), temperature)
+)
+
+# Derive from expression of Gibbs energy differential
+
+_gibbs_energy_expr = sym_symbols("gibbs_energy", cls=SymFunction)(temperature, pressure)
+
+# Since Gibbs energy is a differentiable function of all its parameters, the can apply
```
blackyblack, line 51:
```suggestion
# Since Gibbs energy is a differentiable function of all its parameters, one can apply
```

PR #710, opened 2024-04-25
Derive: work in isothermal process

alesanter:


symplyphysics/laws/thermodynamics/work_of_ideal_gas_in_isothermal_process.py
```diff
@@ -30,9 +43,39 @@
...
+    ideal_gas_equation.mole_count: _mole_count,
+    ideal_gas_equation.temperature: temperature,
+})
+
+_pressure_expr = solve(_ideal_gas_eqn, ideal_gas_equation.pressure)[0]
+
+_volume_before = SymSymbol("volume_before", positive=True)
+_volume_after = SymSymbol("volume_after", positive=True)
+
+_work_expr = work_law.law.rhs.xreplace({
```
blackyblack, line 65: Use `subs`

symplyphysics/laws/thermodynamics/work_of_ideal_gas_in_isothermal_process.py
```diff
@@ -30,9 +43,39 @@
...
+    work_law.volume: ideal_gas_equation.volume,
```
blackyblack, line 66: Use inverse replacement - substitute `ideal_gas_equation.volume` for `work_law.volume`

symplyphysics/laws/thermodynamics/work_of_ideal_gas_in_isothermal_process.py
alesanter: It doesn't work with `subs`. Seems like the issue is that `volume` is a bound symbol in the integral, and only `xreplace` can be used. But when the integral doesn't have bounds, `subs` can be used, too.

symplyphysics/laws/thermodynamics/work_of_ideal_gas_in_isothermal_process.py
alesanter: That's a good solution.

blackyblack (approved):

PR #730, opened 2024-04-27
Law: general heat equation

alesanter:


test/thermodynamics/heat_transfer/general_equation_in_one_dimension_test.py
```diff
@@ -0,0 +1,42 @@
...
+
+@fixture(name="test_args")
+def test_args_fixture() -> Args:
+    rho = 1
+    cv = 1
+    k = 1
+    q = 0
+    T = sin(heat_equation.position) * exp(-1 * heat_equation.time)
+    t = 1
+    x = 1
```
blackyblack, line 21: Try to look for some real numbers. This is way too synthetic.

test/thermodynamics/heat_transfer/general_equation_in_one_dimension_test.py
```diff
@@ -0,0 +1,42 @@
...
+    lhs = heat_equation.law.lhs.subs(values)
+    rhs = heat_equation.law.rhs.subs(values)
+    assert expr_equals(lhs, rhs)
+    values = {
+        heat_equation.position: test_args.x,
+        heat_equation.time: test_args.t,
+    }
+    lhs = lhs.subs(values)
+    rhs = rhs.subs(values)
+    assert expr_equals(lhs, rhs)
```
blackyblack, line 42: Now I think this second check does not make much sense.

test/thermodynamics/heat_transfer/general_equation_in_one_dimension_test.py
alesanter: image
The problem is that for more "real" numbers the floating point arithmetic doesn't get on well with `assert_equal`.

alesanter: This gets better if I represent the values as Rationals

test/thermodynamics/heat_transfer/general_equation_in_one_dimension_test.py
alesanter: Agreed

blackyblack (approved):

PR #743, opened 2024-05-03
Law: time-dependent via time-indepenent solution of Schrödinger equation

alesanter:


symplyphysics/laws/quantum_mechanics/schrodinger/time_dependent_via_time_independent_solution.py
```diff
@@ -0,0 +1,88 @@
...
+)
+
+# Show that the normalization condition still holds
+
+_condition = condition.normalization_condition.subs({
+    condition.position: position,
+    condition.time: time,
+})
+
+_tise_condition = _condition.replace(
```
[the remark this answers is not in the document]
alesanter: For functions `replace` works every time unlike `subs` and it's generally more convenient because I don't need to think about function arguments, `sympy` will replace everything under the hood itself

test/quantum_mechanics/schrodinger/time_dependent_via_time_independent_solution_test.py
```diff
@@ -24,7 +24,7 @@ def test_args_fixture() -> Args:
 
 def test_law(test_args: Args) -> None:
     result = law.calculate_time_dependent_wave_function_value(test_args.psi, test_args.e, test_args.t)
-    assert_equal(result, (-3.00-1.02j) / sqrt(units.meter), tolerance=3e-3)
+    assert_equal(result, (-0.844-3.05j) / sqrt(units.meter))
```
blackyblack, line 27: Why it's different from previous value?

alesanter: Because `hbar.scale_factor` is not the same as its SI value

blackyblack: Your initial number was wrong?

alesanter: I'm guessing I had an error in my initial calculations

blackyblack (approved):

PR #745, opened 2024-05-04
Plot: energy eigenfunctions of quantum harmonic oscillator

alesanter:


plots/laws/quantum_mechanics/harmonic_oscillatior_eigenfunctions.py
```diff
@@ -0,0 +1,44 @@
...
+    backend=MatplotlibBackend,
+    legend=True,
+    show=False,
+    annotations=None,
+)
+
+law = wave_eigenfunctions.law.rhs.subs({
+    pi: pi.evalf(),
+    wave_eigenfunctions.hbar: convert_to(units.hbar, units.joule * units.second).evalf(),
+}).evalf()
```
blackyblack, line 29: Why this `evalf`?

blackyblack (approved):

plots/laws/quantum_mechanics/harmonic_oscillatior_eigenfunctions.py
alesanter: At first there were some problems with lambdifying expressions so I `evalf`ed everything. But looks like it's fine without it too

PR #754, opened 2024-05-07
Law: probability of macrostate

alesanter:


symplyphysics/laws/thermodynamics/probability_of_ideal_gas_macrostate.py
```diff
@@ -0,0 +1,94 @@
...
+## The _macrostate_ of the gas only requires the description of the particle count in each cell of the vessel.
+## If we assume the particles to be identical, it doesn't make a difference, from a macroscopic standpoint, particles
+## of which number appear in which cells. Hence the probability of the macrostate is larger than that of the microstate
+## by the factor of the number of permutations of the particles among the cells, which is exactly the statistical weight
+## of the macrostate.
+
+# Law: P = G * Product(p_i**N_i, i)
+## P - probability of macrostate
+## G - statistical weight of the macrostate
+## p_i - probability of finding one particle in i-th cell
```
blackyblack, line 35: Exactly one or at least one?

symplyphysics/laws/thermodynamics/probability_of_ideal_gas_macrostate.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@validate_input(
+    one_particle_in_cell_probabilities_=one_particle_in_cell_probability,
+    particle_count_in_cells_=particle_count_in_cell,
+)
+@validate_output(macrostate_probability)
+def calculate_macrostate_probability(
+    one_particle_in_cell_probabilities_: Sequence[Probability],
+    particle_count_in_cells_: Sequence[int],
```
blackyblack, line 66: Might be a good idea to combine them in one list: `Sequence[Tuple[Probability, int]]`

symplyphysics/laws/thermodynamics/probability_of_ideal_gas_macrostate.py
```diff
@@ -0,0 +1,94 @@
...
+        raise ValueError("The number of particle counts must equal the number of probabilities")
+
+    try:
+        assert_equal(sum(one_particle_in_cell_probabilities_), 1)
+    except AssertionError as e:
+        raise ValueError("The probabilities must sum up to 1") from e
+
+    statistical_weight_ = factorial(sum(particle_count_in_cells_))
+    for particle_count_ in particle_count_in_cells_:
+        statistical_weight_ = statistical_weight_ // factorial(particle_count_)
```
blackyblack, line 78: You could simply pass statistical weight here, instead of calculating it. Let's keep it this way, but put a link to a law, that justifies this calculation.

test/thermodynamics/probability_of_ideal_gas_macrostate_test.py
```diff
@@ -0,0 +1,53 @@
...
+    p3 = Probability(0.5)
+    n1 = 10
+    n2 = 13
+    n3 = 5
+    return Args(p1=p1, p2=p2, p3=p3, n1=n1, n2=n2, n3=n3)
+
+
+def test_law(test_args: Args) -> None:
+    ps = test_args.p1, test_args.p2, test_args.p3
+    ns = test_args.n1, test_args.n2, test_args.n3
```
blackyblack, line 23: Better use lists instead of tuples: `ns = [test_args.n1, test_args.n2, test_args.n3]`

blackyblack (approved):

symplyphysics/laws/thermodynamics/probability_of_ideal_gas_macrostate.py
alesanter: I thought about it, but in my opinion this places an additional burden on the user to pre-calculate it. Maybe we can use an [already written function](https://github.com/blackyblack/symplyphysics/blob/0ca92cbcedea44b3388e78f5207f2e321780fa9f/symplyphysics/laws/thermodynamics/maxwell_boltzmann_statistics/statistical_weight_of_macrostate.py)?

symplyphysics/laws/thermodynamics/probability_of_ideal_gas_macrostate.py
alesanter: Fair point, at least one

symplyphysics/laws/thermodynamics/probability_of_ideal_gas_macrostate.py
blackyblack: Better not. We tend to use loose coupling between the laws. So we only import other law, when current law depends on it, ie when deriving one law from another. Here, to maintain loose coupling we either explain, what we are doing here. Or accept statistical weight as parameter and let user decide, how to calculate it. First option looks better.

symplyphysics/laws/thermodynamics/probability_of_ideal_gas_macrostate.py
```diff
@@ -4,13 +4,13 @@
...
     Symbol,
     convert_to_float,
-    validate_input,
     validate_output,
     ProductIndexed,
     SymbolIndexed,
     global_index,
     assert_equal,
 )
+from symplyphysics.core.quantity_decorator import _assert_expected_unit
```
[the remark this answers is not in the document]
alesanter: I know that. It's only now that I remembered we have `assert_equivalent_dimension`.

blackyblack (approved):

PR #771, opened 2024-05-15
Law: relativistic addition of parallel velocities

alesanter:


symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
```diff
@@ -0,0 +1,70 @@
...
+    vector_magnitude,
+)
+
+# Description
+## Consider two inertial reference frames: one fixed (lab frame) and one tied to the moving object (proper frame).
+## The proper frame is moving with some velocity `v` relative to lab frame. Then, according to the theory of special
+## relativity, the velocity of the object relative to lab frame is not equal to the sum of its velocity in the proper
+## frame and the velocity of the proper frame relative to the lab frame.
+
+# Law: u_parallel = (u_parallel' + v) / (1 + dot(u_parallel', v) / c**2)
```
blackyblack, line 22: Now it's quite tricky to derive `u_parallel'` from this law. Might be a good idea to have both this and inverted forms of the law.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
```diff
@@ -0,0 +1,70 @@
...
+## u_parallel - velocity vector relative to lab frame in the perpendicular direction
```
blackyblack, line 23: Cannot understand, what is perpendicular here. And why it is called `parallel` if it's in perpendicular direction.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
```diff
@@ -0,0 +1,70 @@
...
+## u_parallel' - velocity vector relative to proper frame in the perpendicular direction
+## v - velocity vector of proper frame relative to lab frame
+## c - speed of light
+
+
+def parallel_addition_law(
```
blackyblack, line 29: Rename to make clear, what the result of addition is.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
```diff
@@ -0,0 +1,70 @@
...
+
+    factor = 1 + dot / speed_of_light**2
+
+    added = add_cartesian_vectors(
+        parallel_velocity_component_in_proper_frame_,
+        proper_frame_velocity_,
+    )
+
+    scaled = scale_vector(1 / factor, added)
+    return scaled
```
blackyblack, line 46: `scaled` is unused here. Simply return the result: `return scale_vector(1 / factor, added)`

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
```diff
@@ -0,0 +1,70 @@
...
+    parallel_velocity_component_in_proper_frame_: QuantityVector,
+    proper_frame_velocity_: QuantityVector,
+) -> QuantityVector:
+    cross = cross_cartesian_vectors(
+        parallel_velocity_component_in_proper_frame_.to_base_vector(),
+        proper_frame_velocity_.to_base_vector(),
+    )
+
+    if Quantity(vector_magnitude(cross)).scale_factor != 0:
+        raise ValueError("The two velocity vectors must be collinear")
```
blackyblack, line 64: If they should be collinear, what's the reason to have this law in vector form? Simply use scalar addition and multiplication.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
```diff
@@ -0,0 +1,70 @@
...
+def calculate_parallel_velocity_component_in_lab_frame(
+    parallel_velocity_component_in_proper_frame_: QuantityVector,
+    proper_frame_velocity_: QuantityVector,
+) -> QuantityVector:
+    cross = cross_cartesian_vectors(
+        parallel_velocity_component_in_proper_frame_.to_base_vector(),
+        proper_frame_velocity_.to_base_vector(),
+    )
+
+    if Quantity(vector_magnitude(cross)).scale_factor != 0:
```
blackyblack, line 63: With real numbers this check will never pass. Might be a good idea to use approximate equality check.

test/relativistic/vector/relativistic_velocity_parallel_to_movement_test.py
```diff
@@ -0,0 +1,72 @@
...
+    assert_equal,
+    units,
+    Quantity,
+    QuantityVector,
+    errors,
+)
+from symplyphysics.laws.relativistic.vector import relativistic_velocity_parallel_to_movement as law
+
+# Description
+## A body is moving parallel to a frame of reference that is moving relative to the laboratory frame of reference.
```
blackyblack, line 14: I think we decided that body is attached to a proper frame. So body is moving in the frame of reference.

blackyblack, line 14: If we say, that body is moving in a frame of reference, what is it moving parallel to? Probably to the lab frame, I guess.

test/relativistic/vector/relativistic_velocity_parallel_to_movement_test.py
```diff
@@ -0,0 +1,72 @@
...
+def test_args_fixture() -> Args:
+    u = QuantityVector([Quantity(0.1 * speed_of_light), Quantity(0.2 * units.speed_of_light), 0])
+    v = QuantityVector([Quantity(-0.2 * speed_of_light), Quantity(-0.4 * units.speed_of_light), 0])
+    return Args(u=u, v=v)
+
+
+def test_law(test_args: Args) -> None:
+    result = law.calculate_parallel_velocity_component_in_lab_frame(test_args.u, test_args.v)
+    assert len(result.components) == 3
+    for result_component, correct_value in zip(result.components, [-0.111, -0.222, 0]):
```
blackyblack, line 32: Why not use `1 / 9`, '2 / 9' like in description?

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
alesanter: You can get `u_parallel'` by replacing `v` with `-v`. I'll write it in the `Notes`

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
alesanter: Maybe just name such functions just `law`? The content of the law should evident from the module name

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
alesanter: It's quite often that we split vectors into parallel and perpendicular components in some problems. I don't see a point in doing more work converting vectors to scalars and the resulting scalar back into a vector.

test/relativistic/vector/relativistic_velocity_parallel_to_movement_test.py
alesanter: I reworded it

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
blackyblack: Functions are usually named as verbs, i.e. `it_does_something()`. And if we are looking to get `u_parallel'` or `v` from this law, we would need 2 additional functions.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
blackyblack: From vector form we are unable to solve for another parameter of a law, i.e. obtain `v`. We should use vector form for non collinear vectors.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
```diff
@@ -0,0 +1,77 @@
...
+)
+from symplyphysics.core.approx import APPROX_RELATIVE_TOLERANCE
+
+# Description
+## Consider two inertial reference frames: one fixed (lab frame) and one tied to the moving object (proper frame).
+## The proper frame is moving with some velocity `v` relative to the lab frame. Then, according to the theory of special
+## relativity, the velocity of the object relative to lab frame is not equal to the sum of its velocity in the proper
+## frame and the velocity of the proper frame relative to the lab frame.
+
+# Law: u_parallel = (u_parallel' + v) / (1 + dot(u_parallel', v) / c**2)
```
blackyblack, line 24: It seems, it is limited to special relativity. Add to conditions, if I am correct.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
```diff
@@ -0,0 +1,77 @@
...
+@validate_input(
+    parallel_velocity_component_in_proper_frame_=units.velocity,
+    proper_frame_velocity_=units.velocity,
+)
+@validate_output(units.velocity)
+def calculate_parallel_velocity_component_in_lab_frame(
+    parallel_velocity_component_in_proper_frame_: QuantityVector,
+    proper_frame_velocity_: QuantityVector,
+    *,
+    tolerance_: float = APPROX_RELATIVE_TOLERANCE,
```
blackyblack, line 64: I think, `assert_equal` should use this default value, if `None` is passed. Give it a try.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
alesanter: You're right

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
alesanter: Why would we need to solve for `v`? It's quite often the case that not all symbols are solvable for in vector equations.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
alesanter: I just thought that this is somewhat inconsistent with the fact that we have `law`s in scalar laws.

We can reuse the same function to find `u_paralell'`. `proper` and `lab` frames switch, the only difference is that `v` becomes `-v`. I don't really think we need a separate function for that, it's redundant from the physical standpoint.

And I'm not sure there is a formula for `v`.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
alesanter: No, it passes None into the `assert_equal` and this yield an error.

alesanter: We could make the default `tolerance` value in `assert_equal` a `None` and then set it to `APPROX_RELATIVE_TOLERANCE` in the function body.

blackyblack: We could, but it should be done by default parameters, not manually. I will check, what is the proper way of passing default parameters for a function.

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
blackyblack: Yes, we could do it. But with `solve` we could get solution for `u_parallel'` automatically and fool proof. It is easy to make a mistake when substituting it manually. Also requires to thoroughly read the description, or know in advance, how to make the translation.

alesanter: Added them

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
alesanter: You're right

blackyblack (approved):

symplyphysics/laws/relativistic/vector/relativistic_velocity_parallel_to_movement.py
blackyblack: I've fixed it. Take a look, how it can be done.
Next time avoid fancy parameters in test methods. It does not pay off. We try to make core library as flexible and robust as possible. But laws are illustrative mostly, especially `calculate` functions. Keep them simple.

PR #772, opened 2024-05-15
Law: Lorentz transformation of time

alesanter:


symplyphysics/laws/relativistic/lorentz_transformation_of_time.py
```diff
@@ -0,0 +1,72 @@
...
+## clocks that are synchronized according to the [Einstein rule](https://en.wikipedia.org/wiki/Einstein_synchronisation).
+## Let `x, y, z, t` be the coordinates and time of some event in frame `S`, and `x', y', z', t'` be the coordinates
+## and time of the same event in frame `S'`. Assuming that the space is uniform and isotropic and that the time is
+## uniform, there exists a linear dependence between `x, y, z, t` and `x', y', z', t'`, which is called the Lorentz
+## transformation of space and time.
+
+# Law: t' = (t - (V * x) / c**2) / sqrt(1 - (V / c)**2)
+## t' - time in proper (S') reference frame
+## t - time in lab (S) reference frame
+## x - position in lab (S) reference fram
```
blackyblack, line 24:
```suggestion
## x - position in lab (S) reference frame
```

symplyphysics/laws/relativistic/lorentz_transformation_of_time.py
```diff
@@ -0,0 +1,72 @@
...
+# Description
+## Let us consider two inertial reference frames `S` (lab frame) and `S'` (proper frame). The proper frame moves
+## with speed `v` relative to the lab frame. Both frames contain identical fixed (relative to their own frame)
+## clocks that are synchronized according to the [Einstein rule](https://en.wikipedia.org/wiki/Einstein_synchronisation).
+## Let `x, y, z, t` be the coordinates and time of some event in frame `S`, and `x', y', z', t'` be the coordinates
+## and time of the same event in frame `S'`. Assuming that the space is uniform and isotropic and that the time is
+## uniform, there exists a linear dependence between `x, y, z, t` and `x', y', z', t'`, which is called the Lorentz
+## transformation of space and time.
+
+# Law: t' = (t - (V * x) / c**2) / sqrt(1 - (V / c)**2)
```
blackyblack, line 21: It's a bit weird that time dilation depends on position in stationary frame.

symplyphysics/laws/relativistic/lorentz_transformation_of_time.py
```diff
@@ -0,0 +1,72 @@
...
+## - In this law, the Lorentz transformation from the lab frame `S` into the proper frame `S'` is described. In order to get
+##   an opposite transformation (from the proper frame `S'` into the lab frame `S`), replace all primed variables with unprimed
+##   ones and vice verce, and replace `V` with `-V`. This is consistent with the fact that frame `S` can be viewed as moving
+##   with speed `-V` relative to frame `S'`, and hence the same Lorentz transformation can be applied.
+## - In the limit of `V/c << 1` the formula reduces to the classical Galilean transformation `t' = t`.
+
+time_in_proper_frame = Symbol("time_in_proper_frame", units.time)
+time_in_lab_frame = Symbol("time_in_lab_frame", units.time)
+position_in_lab_frame = Symbol("position_in_lab_frame", units.length)
+relative_frame_speed = Symbol("relative_frame_speed", units.velocity)  # speed of proper frame relative to lab frame
```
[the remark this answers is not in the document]
alesanter: I think we've got to adhere to the already established naming conventions. After all, this is a _physics_ related library, and I personally would expect more well known quantity names.
image
image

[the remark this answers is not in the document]
blackyblack: Yes, I agree to it. But we tend to use descriptive names, not conventional. However it seems proper frame is not just any moving frame.
I suggest to keep using proper, rest and lab frames, but add their description somewhere.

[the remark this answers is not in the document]
alesanter: In this very case the problem seems to me to be more of understanding these descriptions. What is `stationary` in `stationary_frame`, the frame or the body? The same thing with the `moving_frame`. It's usually the object that is stationary in the `moving_frame`, so this creates kind of a confusion to me. We can explain the names of the variables in the description.

symplyphysics/laws/relativistic/lorentz_transformation_of_time.py
alesanter: Well, it's just how physics works. This is ultimately the manifestation of the fact that space and time are interconnected. And there's a formal derivation of the Lorentz transformation, if it still bugs you.

symplyphysics/laws/relativistic/lorentz_transformation_of_time.py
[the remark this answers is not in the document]
blackyblack: Yes please.

symplyphysics/laws/relativistic/lorentz_transformation_of_time.py
blackyblack: It's not like I doubt this law is correct.
The reason for my confusion is that this law describes a conversion of 4-dimensional point from one frame to another. And here law looks like it describes time flow in another reference frame.
I suggest to somehow explain that this law is about one component of a point. Maybe even put all 4 components in one law.
There is also an idea of making it not a law, but coordinate transformation function in the `core`.

alesanter: Yes, these are in fact 4-vectors and they require separate machinery. Would be a good thing to implement them later on.

blackyblack (approved):

PR #776, opened 2024-05-15
Law: proper time

alesanter:


symplyphysics/laws/relativistic/proper_time_for_timelike_intervals.py
```diff
@@ -0,0 +1,43 @@
...
+    assert_equal,
+)
+
+# Description
+## In relativity, proper time along a timelike world line is defined as the time as measured by a clock following
+## that line. The proper time interval between two events on a world line is the change in proper time, which is 
+## independent of coordinates, and is a Lorentz scalar.
+
+# Law: Δτ = Δs / c
+## Δτ - proper time
```
blackyblack, line 18: It's time interval, not time.

symplyphysics/laws/relativistic/proper_time_for_timelike_intervals.py
```diff
@@ -0,0 +1,43 @@
...
+## Δs - spacetime interval between two events
+## c - speed of light
+
+# Conditions
+## - The interval is timelike, i.e. `Δs` is real
+
+proper_time = Symbol("proper_time", units.time, real=True)
+spacetime_interval = Symbol("spacetime_interval", units.length)
```
blackyblack, line 26: It should be real as well, according to conditions.

symplyphysics/laws/relativistic/proper_time_for_timelike_intervals.py
```diff
@@ -0,0 +1,43 @@
...
+
+law = Eq(proper_time, spacetime_interval / speed_of_light)
+
+
+@validate_input(spacetime_interval_=spacetime_interval)
+@validate_output(proper_time)
+def calculate_proper_time(
+    spacetime_interval_: Quantity,
+) -> Quantity:
+    _, imag = spacetime_interval_.scale_factor.as_real_imag()
```
blackyblack, line 36: Did you try `Im(spacetime_interval_)`?

symplyphysics/laws/relativistic/proper_time_for_timelike_intervals.py
```diff
@@ -0,0 +1,43 @@
...
+    try:
+        assert_equal(imag, 0)
+    except AssertionError as e:
+        raise ValueError("Interval must be real") from e
```
blackyblack, line 40: Don't try and catch. Simply throw when imaginary part is not zero.

test/relativistic/proper_time_for_timelike_intervals_test.py
```diff
@@ -0,0 +1,31 @@
...
+    ds = Quantity(1 * units.millimeter)
+    return Args(ds=ds)
+
+
+def test_law(test_args: Args) -> None:
+    result = time_law.calculate_proper_time(test_args.ds)
+    assert_equal(result, 3.34 * prefixes.pico * units.second, tolerance=2e-3)
+
+
+def test_bad_interval() -> None:
```
blackyblack, line 26: Add test for complex interval.

symplyphysics/laws/relativistic/proper_time_for_timelike_intervals.py
alesanter: Is it because it's thought as a bad coding practice or it's just not necessary?

blackyblack: Former. We catch errors when dealing with external libraries. When working with own code, organizing code flow with exceptions is bad practice. So you should simply check:
```
if Im(value) != 0:
    raise ValueError()
```

symplyphysics/laws/relativistic/proper_time_for_timelike_intervals.py
```diff
@@ -33,11 +33,7 @@
 def calculate_proper_time(
     spacetime_interval_: Quantity,
 ) -> Quantity:
-    _, imag = spacetime_interval_.scale_factor.as_real_imag()
-    try:
-        assert_equal(imag, 0)
-    except AssertionError as e:
-        raise ValueError("Interval must be real") from e
+    assert_equal(im(spacetime_interval_.scale_factor), 0)
```
blackyblack, line 36: Don't use `assert_equal`. It checks with tolerance and here it should be strictly zero to pass as real number.

test/relativistic/proper_time_for_timelike_intervals_test.py
```diff
@@ -0,0 +1,36 @@
...
+    sb = Quantity(1 * units.coulomb)
+    with raises(errors.UnitsError):
+        time_law.calculate_proper_time(sb)
+    with raises(TypeError):
+        time_law.calculate_proper_time(100)
+
+    sb = Quantity((1 + 2 * I) * units.meter)
+    with raises(ValueError):
```
blackyblack, line 35: It does not pass.

symplyphysics/laws/relativistic/proper_time_for_timelike_intervals.py
alesanter: Yeah, I see. But due to rounding errors the imaginary part might not be exactly zero, right? That's why I wanted to use `assert_equal`. Perhaps, if this function worked more like `expr_equals`, which returns a `bool` we then plug into `assert`, we could use it here.

blackyblack: I think we'll usually work in real numbers space or complex numbers space. When dealing with real numbers, we are unlikely to get imaginary part due to rounding errors. That's why I suggest to use exact equality.

blackyblack (approved):

symplyphysics/laws/relativistic/proper_time_for_timelike_intervals.py
```diff
@@ -0,0 +1,38 @@
...
+
+law = Eq(proper_time, spacetime_interval / speed_of_light)
+
+
+@validate_input(spacetime_interval_=spacetime_interval)
+@validate_output(proper_time)
+def calculate_proper_time(
+    spacetime_interval_: Quantity,
+) -> Quantity:
+    assert im(spacetime_interval_.scale_factor) == 0
```
blackyblack, line 35: Don't use `assert`. Throw `ValueError`.

PR #778, opened 2024-05-15
Plot: Lorentz transform of coordinate

alesanter:


symplyphysics/laws/relativistic/coordinate_conversion_velocity_constant.py
```diff
@@ -15,7 +14,7 @@
 ## Then there are simple transformations that can be used to get the x coordinate in one frame of reference,
 ## knowing the x coordinate in another frame of reference.
 
-## Law is: x_2 = (x_1 - v * t) / (1 - (v / c)^2), where
+## Law is: x_2 = (x_1 - v * t) / sqrt(1 - (v / c)^2), where
```
blackyblack, line 17: Hm... Why such mistake was not caught by the test?

alesanter: The `sqrt` is present in the calculation function. It's only the description that was wrong.

blackyblack (approved):

PR #783, opened 2024-05-18
Example: free electron's light radiation

alesanter:


examples/relativistic/possibility_of_light_radiation_of_free_electron.py
```diff
@@ -0,0 +1,86 @@
...
+energy_conservation_eqn = energy_conservation_eqn.replace(lorentz_factor_expr, lorentz_factor)
+
+print("Energy conservation equation:")
+print(print_expression(energy_conservation_eqn))
+
+# We know that the Lorentz factor is always greater than 1, so we can use that information to figure
+# out if the above equation ever holds true.
+
+electron_rest_mass_range = reduce_inequalities(
+    energy_conservation_eqn & (lorentz_factor > 1),
```
blackyblack, line 69: Another way to deal with it, is to solve for positive `t`, where `t = lorentz_factor - 1`

examples/relativistic/possibility_of_light_radiation_of_free_electron.py
```diff
@@ -0,0 +1,86 @@
...
+    [electron_rest_mass],
+)
+
+# The above inequality never holds, therefore `reduce_inequalities` returns `False`
+assert electron_rest_mass_range == False
```
blackyblack, line 74:
```suggestion
assert electron_rest_mass_range is False
```

examples/relativistic/possibility_of_light_radiation_of_free_electron.py
```diff
@@ -0,0 +1,86 @@
...
+
+# This indicates that free electrons are unable to radiate photons. Only accelerated electrons can do that,
+# for example, when the electron is in the vicinity of the atomic nucleus or is interacting with the electric
+# field.
+
+# What is more, we can use the same expressions that we derived here to show that free electrons cannot
+# absorb light either: `total_energy_after_radiation` gives the energy of the electron + photon system before
+# the absorption has happened in a reference frame where the total momentum of the system is zero, and
+# `total_energy_before_radiation` gives the energy of the electron after photon absorption.
```
blackyblack, line 83: I would love to see it as code, instead of explanation.

examples/relativistic/possibility_of_light_radiation_of_free_electron.py
```diff
@@ -0,0 +1,86 @@
...
+
+# Note that this result is only true for elementary particles and doesn't hold for compound objects consisting
+# of many elementary particles. These can absorb and radiate light freely, without the need to be accelerated.
```
blackyblack, line 86: Could you briefly explain this effect? On first sight it looks like the same proof works for any particle or set of particles.

examples/relativistic/possibility_of_light_radiation_of_free_electron.py
alesanter: `sympy` returns its own version of `False` so this would not be correct

examples/relativistic/possibility_of_light_radiation_of_free_electron.py
alesanter: I think the way with `reduce_inequalities` less obscure. And actually `sympy` can figure out that its always false even without solving for electron's mass.

examples/relativistic/possibility_of_light_radiation_of_free_electron.py
```diff
@@ -62,25 +62,37 @@
...
-# What is more, we can use the same expressions that we derived here to show that free electrons cannot
-# absorb light either: `total_energy_after_radiation` gives the energy of the electron + photon system before
-# the absorption has happened in a reference frame where the total momentum of the system is zero, and
-# `total_energy_before_radiation` gives the energy of the electron after photon absorption.
+# Let us now consider the case of photon getting absorbed by the electron. We assume the reference frame
+# where the total momentum of the electron + photon system is zero.
+
+total_energy_before_absorption = moving_electron_energy + photon_energy
+
+# The total momentum is zero before and after absorption, therefore the electron is at rest after absorption.
```
blackyblack, line 80: Not formally justified. We could prove it formally, but it is not necessary for an example.
Still, some additional description would be handy. I.e. momentum before absorption equals to momentum after, hence momentum after is zero. After absorption there is only electron left, so its momentum is zero. Hence its velocity is zero for non-zero mass.

blackyblack (approved):

examples/relativistic/possibility_of_light_radiation_of_free_electron.py
blackyblack: Well, pylint does not like this.
And it is generally bad coding practice to compare with True/False. I.e. we write `if some_expression` instead of `if some_expression is True`.
Pylint would be satisfied with this approach:
```
assert not electron_rest_mass_range
```
but it does not look quite good.

alesanter: Not sure there is a better approach to it. We need to checks that the result is strictly `False` which indicates that the reduction has failed. But I see, talking about the `True/False` thing.

blackyblack: Let's keep it `== False`. It's SymPy bad design that led to such Pylint warning. Proper design was to return empty list `[]` for reduced inequalities range.

alesanter: Actually we could import `false` from `sympy` and do `reduction is false`

blackyblack (approved):

PR #797, opened 2024-05-22
Law: kinetic energy via speed and momentum

alesanter:


symplyphysics/laws/dynamics/kinetic_energy_differential.py
```diff
@@ -0,0 +1,45 @@
...
+## d(x) - exact differential of `x`
+
+# Notes
+## - Works in the case of special relativity as well
+
+kinetic_energy_differential = Symbol("kinetic_energy_differential", units.energy)
+speed = Symbol("speed", units.velocity)
+momentum_differential = Symbol("momentum", units.momentum)
+
+law = Eq(kinetic_energy_differential, speed * momentum_differential)
```
blackyblack, line 27: Why not `Derivative` ?

blackyblack: Let's use `Derivative`

blackyblack (approved):

PR #799, opened 2024-05-23
Example: relativistic movement due to constant force

alesanter:


examples/relativistic/constant_force_trajectory.py
```diff
@@ -0,0 +1,85 @@
...
+print(print_expression(Eq(speed, speed_via_time)))
+
+# Integrate the velocity formula over time to obtain the distance formula.
+# Note that the distance covered is zero at `t = 0`.
+
+distance, length_term = symbols("distance, length_term", positive=True)
+
+distance_via_time = speed_via_time.integrate(time)
+
+length_term_expr = distance_via_time.subs(time, 0)
```
blackyblack, line 55: But distance at zero time is zero. Why you do not use this condition?

examples/relativistic/constant_force_trajectory.py
```diff
@@ -0,0 +1,85 @@
...
+    ),
+    (distance, rest_mass),
+    dict=True,
+)[0][distance]
+
+print("\n\nFormula of particle distance as a function of time:")
+print(print_expression(Eq(distance, distance_via_time)))
+
+print("\n\nFormula of `length_term`:")
+print(print_expression(length_term_expr))
```
blackyblack, line 70: What is `length_term` and why do we care about it?

examples/relativistic/constant_force_trajectory.py
alesanter: I renamed the symbols so that they're less confusing.

blackyblack: Still, this integration constant should be zero in our example.

alesanter: Why? When we integrate speed over time we do not get distance immediately. We obtain some function of time that is not necessarily zero at `t = 0`.

blackyblack: You even have a note about it:
`Note that the distance covered is zero at t = 0`

blackyblack: Yes, but why not simply equate it to zero and get the same distance equation?

blackyblack: It's weird that SymPy `integrate()` does not add integration constant to the result.
For your example it should be `x(t) = -cos(t) + C1` after integration, where `C1` is initial position.

blackyblack: From SymPy documentation:
```
Note that SymPy does not include the constant of integration. If you want it, you can add one yourself, or rephrase your problem as a differential equation and use dsolve to solve it, which does add the constant (see [Solving Differential Equations](https://docs.sympy.org/latest/tutorials/intro-tutorial/solvers.html#tutorial-dsolve)).
```
So you effectively wrote a workaround for this issue, by subtracting constant term from the result.

examples/relativistic/constant_force_trajectory.py
```diff
@@ -0,0 +1,85 @@
...
+
+print("Formula of particle speed as a function of time:\n")
+print(print_expression(Eq(speed, speed_via_time)))
+
+# Integrate the velocity formula over time to obtain the distance formula.
+# Note that the distance covered is zero at `t = 0`.
+
+distance, integration_constant = symbols("distance, integration_constant", positive=True)
+
+speed_integrated_over_time = speed_via_time.integrate(time)
```
blackyblack, line 53: Let's use descriptive names. Speed integrated over time is position of particle.

blackyblack (approved):

PR #802, opened 2024-05-24
Law: vector of relativistic momentum

alesanter:


symplyphysics/laws/relativistic/vector/relativistic_momentum.py
```diff
@@ -0,0 +1,54 @@
...
+rest_mass = Symbol("rest_mass", units.mass)
+
+
+def momentum_law(velocity_: Vector) -> Vector:
+    v_dot_v = dot_vectors(velocity_, velocity_)
+    factor = rest_mass / sqrt(1 - v_dot_v / speed_of_light**2)
+    return scale_vector(factor, velocity_)
+
+
+def velocity_law(momentum_: Vector) -> Vector:
```
blackyblack, line 35: Maybe also add law for rest mass?

blackyblack (approved):

symplyphysics/laws/relativistic/vector/relativistic_momentum.py
alesanter: One can use the scalar law for that

blackyblack: Add to description.

PR #809, opened 2024-05-26
Plot: relativistic force for different acceleration configurations

alesanter:


plots/laws/relativistic/relativistic_force_for_different_acceleration_configurations.py
```diff
@@ -0,0 +1,49 @@
...
+        label=r"$a_{\perp} = 0$",
+        acceleration=Vector([10, 0, 0]),
+    ),
+    Datum(
+        label=r"$a_{\parallel} = 0$",
+        acceleration=Vector([0, 10, -20]),
+    ),
+    Datum(
+        label=r"$a_{\perp} \ne 0, a_{\parallel} \ne 0$",
+        acceleration=Vector([10, 10, -20]),
```
blackyblack, line 22: I suppose these vectors should have the same magnitude, to make this plot somehow informative.

alesanter: This makes sense. I'll do that

blackyblack (approved):

PR #811, opened 2024-05-26
Law: velocity in lab frame orthogonal to relative frame speed in SR

alesanter:


test/relativistic/vector/relativistic_velocity_orthogonal_to_movement_test.py
```diff
@@ -0,0 +1,83 @@
...
+    for result_component, correct_component in zip(
+        result.components,
+        [0, 0, ul_orthogonal.components[2]],
+        strict=True
+    ):
+        assert_equal(result_component, correct_component)
+
+
+# TODO: fix `assert_equal` behavior with zeros
+@mark.skip("Fix assert_equal with zeroes")
```
alesanter, line 45: Looks like we need to pass the absolute tolerance besides the relative one because the assertion function fails with the default value of absolute tolerance.

blackyblack: I see. I will try for some nice solution.

test/relativistic/vector/relativistic_velocity_orthogonal_to_movement_test.py
```diff
@@ -0,0 +1,84 @@
...
+
+def test_lab_law(test_args: Args) -> None:
+    result = law.calculate_orthogonal_velocity_component_in_lab_frame(test_args.up, test_args.v)
+    ul_orthogonal = reject_cartesian_vector(
+        test_args.ul.to_base_vector(),
+        test_args.v.to_base_vector()
+    )
+    for result_component, correct_component in zip(
+        result.components,
+        [0, 0, ul_orthogonal.components[2]],
```
blackyblack, line 38: Why do you have components 0 and 1 equal to zero?

test/relativistic/vector/relativistic_velocity_orthogonal_to_movement_test.py
```diff
@@ -0,0 +1,84 @@
...
+    v = QuantityVector([Quantity(-0.200 * speed_of_light), Quantity(-0.400 * speed_of_light), Quantity(0)])
+    return Args(up=up, ul=ul, v=v)
+
+
+def test_lab_law(test_args: Args) -> None:
+    result = law.calculate_orthogonal_velocity_component_in_lab_frame(test_args.up, test_args.v)
+    ul_orthogonal = reject_cartesian_vector(
+        test_args.ul.to_base_vector(),
+        test_args.v.to_base_vector()
+    )
```
blackyblack, line 35: In tests we like to see what are the intermediate results. Write `assert` for `ul_orthogonal`

symplyphysics/laws/relativistic/vector/relativistic_velocity_orthogonal_to_movement.py
```diff
@@ -0,0 +1,73 @@
...
+
+def orthogonal_velocity_component_in_lab_frame_law(
+    velocity_in_proper_frame_: Vector,
+    proper_frame_velocity_: Vector,
+) -> Vector:
+    orthogonal_velocity_component_in_proper_frame_ = reject_cartesian_vector(velocity_in_proper_frame_, proper_frame_velocity_)
+
+    lorentz_factor_ = 1 / sqrt(1 - dot_vectors(proper_frame_velocity_, proper_frame_velocity_) / speed_of_light**2)
+
+    scale_factor_ = lorentz_factor_ * (1 + dot_vectors(proper_frame_velocity_, velocity_in_proper_frame_) / speed_of_light**2)
```
blackyblack, line 45: In description there is minus sign: `1 - dot_vectors`

blackyblack, line 45: In description there is parallel component inside the dot product: `dot_vectors(proper_frame_velocity_, velocity_in_proper_frame_projection)`

test/relativistic/vector/relativistic_velocity_orthogonal_to_movement_test.py
alesanter: Well, this was written before we addressed the assertion issue. Now I can write it in a better way.

symplyphysics/laws/relativistic/vector/relativistic_velocity_orthogonal_to_movement.py
alesanter: They yield to the same result and I wanted this law to be similar to the parallel counterpart.

blackyblack: You should either project it or fix description. If you'll project vector, add small comment, that the result is the same, as when using unprojected vector.

test/relativistic/vector/relativistic_velocity_orthogonal_to_movement_test.py
```diff
@@ -0,0 +1,81 @@
...
+    assert_equal_vectors(
+        result,
+        QuantityVector.from_base_vector(ul_orthogonal),
+        tolerance=3e-3,
+        absolute_tolerance=1e-8,
```
blackyblack, line 40: Don't use both tolerances. I will update code to throw, when both are passed.

alesanter: But this is compatible with how `pytest.assert` works. It accepts both values for tolerance and uses the absolute one if it detects zero, otherwise it defaults to the relative one. IMO it's the correct behavior. The problem is that some values in the vector might be zero and others are non-zero. So we need both of them to make it work.

symplyphysics/laws/relativistic/vector/relativistic_velocity_orthogonal_to_movement.py
alesanter: Sure

test/relativistic/vector/relativistic_velocity_orthogonal_to_movement_test.py
blackyblack: Approx is using the same approach and it led to hard to catch errors. I'll think how to make it better.

symplyphysics/laws/relativistic/vector/relativistic_velocity_orthogonal_to_movement.py
```diff
@@ -45,6 +45,8 @@ def orthogonal_velocity_component_in_lab_frame_law(
         lorentz_factor_def.velocity: vector_magnitude(proper_frame_velocity_),
     })
 
+    # Note that `dot(v, u_parallel') = dot(v, u)`, since `dot(v, u_orthogonal) = 0` and the dot product is linear in
```
blackyblack, line 48: You wrote a note but did not project as I suggested. Now your law does not correspond to description.

alesanter: Fixed

symplyphysics/laws/relativistic/vector/relativistic_velocity_orthogonal_to_movement.py
blackyblack: I am still not happy with this approach. In the description above there is parallel vector. As I said before, you should either update law description, or explicitly project vector.

test/relativistic/vector/relativistic_velocity_orthogonal_to_movement_test.py
blackyblack: I will accept it in this PR but I will ban both tolerances later. We will have option to use `relative_tolerance`, `absolute_tolerance` and `absolute_tolerance_if_zero`. The latter option will be possible to use together with `relative_tolerance`.

alesanter: OK, I'll take a look

alesanter: I'll take a look

symplyphysics/laws/relativistic/vector/relativistic_velocity_orthogonal_to_movement.py
alesanter: I went with fixing the description

test/relativistic/vector/relativistic_velocity_orthogonal_to_movement_test.py
alesanter: Okay, this sounds reasonable.

blackyblack (approved):

PR #816, opened 2024-05-28
Law: velocity of transfer between reference frames

alesanter:


symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
```diff
@@ -0,0 +1,64 @@
...
+)
+
+# Description
+## Suppose two reference frames, one of which is fixed (S) and the other one is moving (S'). The movement of
+## a body stationary in moving frame S' due to the movement of the frame itself is called transfer movement.
+## The velocity related to such movement can be called transfer velocity.
+
+# Law: v_tr = v_0 + cross(w, r)
+## v_tr - vector of transfer velocity
+## v_0 - vector of velocity of moving origin O' of frame S' relative to fixed frame S
```
blackyblack, line 20: What is moving origin?

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
```diff
@@ -0,0 +1,64 @@
...
+## w - vector of angular velocity related to rotation of moving frame S' about instantaneous axis
```
blackyblack, line 21: Why does it rotate? In the description only motion was mentioned.

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
```diff
@@ -0,0 +1,64 @@
...
+## r - vector of position relative to moving frame S'
```
blackyblack, line 22: Position of what?

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
```diff
@@ -0,0 +1,64 @@
...
+    cross_cartesian_vectors,
+)
+
+# Description
+## Suppose two reference frames, one of which is fixed (S) and the other one is moving (S'). The movement of
+## a body stationary in moving frame S' due to the movement of the frame itself is called transfer movement.
+## The velocity related to such movement can be called transfer velocity.
+
+# Law: v_tr = v_0 + cross(w, r)
+## v_tr - vector of transfer velocity
```
blackyblack, line 19: Velocity of what?

test/kinematic/vector/velocity_of_transfer_between_reference_frames_test.py
```diff
@@ -0,0 +1,81 @@
...
+)
+from symplyphysics.core.approx import assert_equal_vectors
+from symplyphysics.laws.kinematic.vector import velocity_of_transfer_between_reference_frames as law
+
+Args = namedtuple("Args", "vtr v0 w r")
+
+
+@fixture(name="test_args")
+def test_args_fixture() -> Args:
+    vtr = QuantityVector([3, -3, 6], dimension=units.velocity)
```
blackyblack, line 17: It's a bit harder to understand values this way. Using normal `meter / second` is easier to read.

test/kinematic/vector/velocity_of_transfer_between_reference_frames_test.py
```diff
@@ -0,0 +1,81 @@
...
+    v0 = QuantityVector([1, -1, 4], dimension=units.velocity)
+    w = QuantityVector([1, 0, -1], dimension=units.frequency)
```
blackyblack, line 19: It has the same dimension as frequency, but it's not frequency.

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
alesanter: Motion implies rotation

test/kinematic/vector/velocity_of_transfer_between_reference_frames_test.py
alesanter: What's the point of the `dimension` parameter then? Also, shouldn't it be something like `measurement_unit` for quantities? It's somewhat counterintuitive to use dimension instead of units there. And I believe it's easier to understand this way, since there is much less clutter.

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
alesanter: Every frame has an origin, and since S' is moving, O' is moving too. I don't really know what's unclear here.

test/kinematic/vector/velocity_of_transfer_between_reference_frames_test.py
blackyblack: No, I meant to write `3 * units.meter / units.second`

alesanter: I see, but isn't it repetitive (and error prone) to write it every time? We can factor it out into some `measurement_unit` parameter.

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
blackyblack: It is unclear, how this origin affects the result. As far as I understand, origin is any random point in S'. And every vector in S' should use this point as a starting point.
Add some notes to the description, if I am correct.

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
blackyblack: Yes, but let's not place too much burden on a reader. Mention that we also allow rotational motion in this formula.

test/kinematic/vector/velocity_of_transfer_between_reference_frames_test.py
blackyblack: Yes, a bit cluttered but helps to understand, what we are doing here. You can extract it to a separate variable, ie to `velocity_unit`

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
```diff
@@ -0,0 +1,67 @@
...
+## a body stationary in moving frame S' due to the movement of the frame itself is called transfer movement.
+## The velocity related to such movement is called transfer velocity. For any material point X, its transfer
+## velocity relative to fixed frame S is the sum of the velocity of moving origin O' of frame S' (since S' is
+## moving) and the cross product of the angular velocity of moving frame's rotation and the position vector of X
+## in moving frame S'.
+
+# Law: v_tr = v_0 + cross(w, r)
+## v_tr - vector of transfer velocity of point X relative to fixed frame S.
+## v_0 - vector of velocity of moving origin O' of frame S' relative to fixed frame S
+## w - vector of angular velocity related to rotation of moving frame S' about instantaneous axis
```
blackyblack, line 24: Pseudovector

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
alesanter: Well, first of all, for any reference frame there is only _one_ chosen point of origin, so it's not _any random point_ technically. Moreover, not every vector in a reference frame has to start at that point. But these are just technical things. The main reason for that lies in the derivation of this law and I'm not really sure what note to write here. The choice of the origin affects the result in such a way that if we were to choose a different origin, we would have to re-calculate vectors and it might also move through space with a different velocity, but isn't that obvious?

blackyblack: The problem with it, we have no mention of origin point before. That's why I want to clarify, what is it and how to choose one.
Suppose you are a user of this library and you have to pass origin velocity as parameter. How should you do it? What is origin of falling brick?
So I supposed, the choice of origin point is pretty much random. You choose one fixed point, that is the most convenient.
As a side note, there may be some problems in keeping this point fixed. For example, if the space of reference frame is bending, there is no obvious way of tracking absolute position of a given point.

alesanter: I'll mention the origin in the description. But isn't there only one origin for any reference frame? So defining a reference frame implies a definite origin.

The origin of falling brick is the origin of the coordinate system that we use to find the coordinates of the brick. It's not some arbitrary point in $`\mathbb{R}^3`$, otherwise we wouldn't have any coordinates.

And I think we only deal with flat spacetime within the scope of our library, so the space bending is out of question, although I can mention that.

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
```diff
@@ -0,0 +1,67 @@
...
+
+# Law: v_tr = v_0 + cross(w, r)
+## v_tr - vector of transfer velocity of point X relative to fixed frame S.
+## v_0 - vector of velocity of moving origin O' of frame S' relative to fixed frame S
+## w - pseudovector of angular velocity related to rotation of moving frame S' about instantaneous axis
+## r - vector of position of point X relative to moving frame S'
+
+
+def transfer_velocity_law(
+    origin_velocity_: Vector,
```
blackyblack, line 29: Since origin (and axes) define reference frame, why don't we use `frame_velocity` here, instead of `origin_velocity`. Or `relative_frame_velocity` to be clear.
This way we avoid confusion in understanding, what is `origin velocity` and what is `origin` in general.

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
blackyblack: In law and description we should talk about velocity of a frame relative to another frame. And we should mention origin point as a way of finding this velocity, i.e. taking some fixed point in a frame and measuring its velocity.
Update description and naming according to this suggestion.

symplyphysics/laws/kinematic/vector/velocity_of_transfer_between_reference_frames.py
alesanter: To me it's more confusing what _frame velocity_ is. And it makes little sense from the derivation, where the two origins (the static one and the moving one) are used to obtain this formula.

alesanter: I'll give it a try with `frame` though.

blackyblack (approved):

PR #818, opened 2024-05-28
Law: absolute velocity of arbitrary motion

alesanter:


symplyphysics/laws/kinematic/vector/absolute_velocity_of_arbitrary_motion.py
```diff
@@ -0,0 +1,66 @@
...
+    scale_vector,
+)
+
+# Description
+## Suppose two reference frames, one of which is fixed (S) and the other one is moving (S'). The movement of the
+## body relative to fixed frame S is called absolute movement. The movement of the body relative to moving frame S'
+## is called relative movement. And the movement of the body due to the movement of reference frame S' is called transfer
+## velocity. The absolute velocity is the sum of these two velocities.
+
+# Law: v_abs = v_rel + v_tr
```
blackyblack, line 17: It looks very similar to https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/kinematic/vector/rolling_velocity_is_translational_plus_rotational_velocities.py
Could you generalize this law, and remove redundant law?

blackyblack (approved):

symplyphysics/laws/kinematic/vector/absolute_velocity_of_arbitrary_motion.py
alesanter: I think this one is general enough to encompass that one. We can get rid of that one.

blackyblack: Yes, but I think we can mention, that this law applies to rolling motion.

blackyblack (approved):

PR #815, opened 2024-05-28
electronic_current_in_probe_circuit_in_gas_plasma

Zulusssss:
add law and tests

PR #819, opened 2024-05-29
concentration_of_ions_in_gas_plasma

Zulusssss:
add law and tests

### DIFF
--- a/symplyphysics/laws/chemistry/concentration_of_ions_in_gas_plasma.py
+++ b/symplyphysics/laws/chemistry/concentration_of_ions_in_gas_plasma.py
@@ -1,0 +1,44 @@
+from sympy import Eq, solve, sqrt, pi
+from sympy.physics.units import elementary_charge, electron_rest_mass, boltzmann_constant
+from symplyphysics import (units, Quantity, Symbol, validate_input, validate_output, clone_symbol, symbols)
+
+# Description
+## A gas–discharge plasma is an ionized gas in which the concentrations of positively and negatively charged particles are approximately
+## equal to each other, and the Debye shielding radius is significantly smaller than the characteristic size of the volume in which the
+## ionized gas is located.
+## A probe is an auxiliary metal electrode that is inserted into a plasma volume for its examination. Flat, cylindrical and spherical probes
+## are used. A potential is set relative to one of the electrodes on the probe and the dependence of the incoming current on the value of
+## this potential is removed.
+## By measuring the probe current, the concentration of ions in the plasma can be determined.
+
+## Law is: n = I0 / (S * e * sqrt(k * T / (2 * pi * m))), where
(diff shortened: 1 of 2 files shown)


symplyphysics/laws/chemistry/concentration_of_ions_in_gas_plasma.py
```diff
@@ -0,0 +1,44 @@
...
+# Description
+## A gas–discharge plasma is an ionized gas in which the concentrations of positively and negatively charged particles are approximately
+## equal to each other, and the Debye shielding radius is significantly smaller than the characteristic size of the volume in which the
+## ionized gas is located.
+## A probe is an auxiliary metal electrode that is inserted into a plasma volume for its examination. Flat, cylindrical and spherical probes
+## are used. A potential is set relative to one of the electrodes on the probe and the dependence of the incoming current on the value of
+## this potential is removed.
+## By measuring the probe current, the concentration of ions in the plasma can be determined.
+
+## Law is: n = I0 / (S * e * sqrt(k * T / (2 * pi * m))), where
```
blackyblack, line 14: It's pretty much the same formula, as in https://github.com/blackyblack/symplyphysics/pull/815/files
Let me know, if we have any reason to keep this law. Will close this PR without merging if they are the same.

symplyphysics/laws/chemistry/concentration_of_ions_in_gas_plasma.py
```diff
@@ -0,0 +1,44 @@
...
+## n - concentration of ions in plasma,
+## I - probe current (at a probe potential equal to the plasma potential),
```
blackyblack, line 16: It's `I0` in formula.

blackyblack (changes requested):

symplyphysics/laws/chemistry/concentration_of_ions_in_gas_plasma.py
Zulusssss: > It's pretty much the same formula, as in https://github.com/blackyblack/symplyphysics/pull/815/files Let me know, if we have any reason to keep this law. Will close this PR without merging if they are the same.

Here we calculate the concentration of ions in the plasma, with a potential on the probe equal to the plasma potential. Thus, we do not affect the distribution of ions in the plasma and calculate the concentration of ions about the same as if the probe did not exist at all. This, it seems to me, is not the most obvious fact. But technically yes, this formula can be obtained from a previously added one.

blackyblack: Thanks for the clarification. But I'm not convinced we should keep it. Closing.

blackyblack:
Duplicate of https://github.com/blackyblack/symplyphysics/pull/815

PR #820, opened 2024-05-29
electric_intensity_in_gas_gap_between_two_electrodes

Zulusssss:
add law and tests


symplyphysics/laws/chemistry/electric_intensity_in_gas_gap_between_two_electrodes.py
```diff
@@ -0,0 +1,35 @@
+from sympy import Eq, solve, sqrt
+from symplyphysics import (units, Quantity, Symbol, validate_input, validate_output,)
+
+# Description
+## Consider a two-electrode gas-filled gap with flat electrodes. Let one of the electrodes emit charged particles into a gaseous medium.
+## If the potential collecting these charged particles is set to another electrode, then a current will flow through the gap.
+
+## Law is: E = (3 / 2) * sqrt(x / L) * U / L, where
```
blackyblack, line 8: I do not see any dependence on the medium, as well as on the geometry of electrodes. So it looks like very general law, that applies to any current between electrodes.
However I do not think it holds true, eg for short circuit.
Check if any additional conditions should be applied. And I do not see anything chemistry related in this law.

symplyphysics/laws/chemistry/electric_intensity_in_gas_gap_between_two_electrodes.py
```diff
@@ -0,0 +1,35 @@
...
+voltage_between_electrodes = Symbol("voltage_between_electrodes", units.voltage)
+
+law = Eq(electric_intensity, (3 / 2) * sqrt(coordinate / distance_between_electrodes) * voltage_between_electrodes / distance_between_electrodes)
+
+
+@validate_input(coordinate_=coordinate,
+    distance_between_electrodes_=distance_between_electrodes,
+    voltage_between_electrodes_=voltage_between_electrodes)
+@validate_output(electric_intensity)
+def calculate_electric_intensity(coordinate_: Quantity, distance_between_electrodes_: Quantity,
```
blackyblack, line 27: Probably should check that coordinate is not over distance between electrodes.

symplyphysics/laws/chemistry/electric_intensity_in_gas_gap_between_two_electrodes.py
Zulusssss: > I do not see any dependence on the medium, as well as on the geometry of electrodes. So it looks like very general law, that applies to any current between electrodes. However I do not think it holds true, eg for short circuit. Check if any additional conditions should be applied. And I do not see anything chemistry related in this law.

Yes, there really is no parameter responsible for the gas in the gap between the electrodes, but the formula is valid for the gas gap. It's just that the parameter characterizing the gas is present in the formula for current density, but not in the formula for field strength.

For more information, see page 30:
https://techeiscatel.ru/media/books/LabPlasma.pdf

blackyblack: Ok. Add Poisson equation boundary conditions to the law conditions. Also add to conditions, that there is gas between electrodes.

blackyblack (approved):

PR #853, opened 2024-07-16
Law: linear density

alesanter:


symplyphysics/laws/quantities/quantity_is_linear_density_times_length.py
```diff
@@ -0,0 +1,66 @@
...
+Symbol:
+    X
+"""
+
+linear_density = SymSymbol("linear_density")
+r"""
+Intensive linear density
+
+Symbol:
+    :math:`\lambda_X`
```
blackyblack, line 31: Here write in ascii symbols, so it can be read without greek letters knowledge.
Put latex form in `Latex:` section.

symplyphysics/laws/quantities/quantity_is_linear_density_times_length.py
```diff
@@ -0,0 +1,66 @@
...
+"""
+Length
+
+Symbol:
+    L
+"""
+
+law = Eq(extensive_quantity, linear_density * length)
+r"""
+:math:`X = \lambda_X L`
```
blackyblack, line 44: Put this in `Latex` section. Use ascii symbols here.

symplyphysics/laws/quantities/quantity_is_linear_density_times_length.py
alesanter: Sure

blackyblack (approved):

PR #854, opened 2024-07-17
Law: gravitational shift of spectral lines

alesanter:


symplyphysics/laws/waves/vector/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+Waves (vector laws)
```
blackyblack, line 2:
```suggestion
Waves (Vector)
```

blackyblack (approved):

PR #855, opened 2024-07-17
Example: conical pendulum

alesanter:


examples/dynamics/thin_rod_as_conical_pendulum.py
```diff
@@ -0,0 +1,164 @@
...
+
+element_mass_expr = solve(
+    (whole_rod_mass_eqn, rod_element_mass_eqn),
+    (linear_density_law.linear_density, element_mass),
+    dict=True,
+)[0][element_mass]
+
+# The rod is in equilibrium, so the total torque acting on the rod is zero. There are three forces
+# acting on the rod: the gravity force, the centrifugal force, and the tension force. Since the tension
+# is parallel to the position vector, its torque will be zero.
```
blackyblack, line 98: "its torque will be zero"
Do you have a law for this?

examples/dynamics/thin_rod_as_conical_pendulum.py
```diff
@@ -0,0 +1,164 @@
...
+)
+
+gravity_torque_acting_on_element = torque_def.torque_definition(
+    position_=element_position_vector,
+    force_=gravity_force_acting_on_element,
+)
+
+gravity_torque_acting_on_rod = integrate_cartesian_vector(
+    gravity_torque_acting_on_element.subs(element_length, 1),
+    (distance_to_element, 0, rod_length),
```
blackyblack, line 113: Use `S.Zero` or `sympify` to make mypy happy.

examples/dynamics/thin_rod_as_conical_pendulum.py
```diff
@@ -0,0 +1,164 @@
...
+
+total_torque_acting_on_rod = vector_magnitude(total_torque_vector_acting_on_rod)
+
+equilibrium_eqn = equilibrium_law.law.subs(
+    equilibrium_law.total_torque, total_torque_acting_on_rod
+)
+
+angular_velocity_expr = solve(equilibrium_eqn, angular_velocity)[1]
+
+angular_velocity_expr = refine(
```
blackyblack, line 154: The result contains `element_mass`. I don't think it should be this way.

examples/dynamics/thin_rod_as_conical_pendulum.py
alesanter: The condition I imported is exactly about this.

examples/dynamics/thin_rod_as_conical_pendulum.py
alesanter: Fixed

examples/dynamics/thin_rod_as_conical_pendulum.py
```diff
@@ -53,8 +54,8 @@
 )
 
 # Let us consider a non-inertial reference frame in which the rod is at rest, i.e. it is in equilibrium.
-# The `z` axis is codirectional to the vector of angular velocity and the `y` axis lies in the plane where
-# the rod is oscillating.
+# The `z` axis is codirectional to the vector of angular velocity and the `y` axis is parallel to the radial
+# axis if a cylindrical coordinate system is assumed.
```
blackyblack, line 58: Well, if Z axis is collinear to angular velocity, Y is obviously orthogonal to it, so it should be parallel to radial plane. So you description of Y axis is useless. I suggest to describe origin point instead.

blackyblack (approved):

PR #858, opened 2024-07-17
Law: potential energy of radial motion of planets

alesanter:


symplyphysics/laws/gravity/radial_motion/potential_energy_of_planetary_motion.py
```diff
@@ -0,0 +1,93 @@
+"""
+Potential energy of radial planetary motion
+===========================================
+
+The total energy of the planet can be viewed as the sum of the kinetic and potential energy.
+The potential energy is in turn the sum of the potential energy due to the gravitational interaction
+between the planet and the sun, and the the energy of the tangential motion, which depends on the
```
blackyblack, line 7:
```suggestion
between the planet and the Sun, and the the energy of the tangential motion, which depends on the
```

blackyblack (approved):

PR #885, opened 2024-07-30
Plot: Maxwell—Jüttner distribution

alesanter:


symplyphysics/core/convert.py
```diff
@@ -17,3 +20,30 @@ def convert_to(value: Expr, target_unit: Expr) -> Expr:
...
 def convert_to_float(value: Expr) -> float:
     return float(convert_to(value, S.One))
+
+
+_si_conversions: dict[units.Dimension, Expr] = {
+    dimension_definitions.angle: S.One,
+    dimension_definitions.length: units.meter,
+    dimension_definitions.mass: units.kilogram,
+    dimension_definitions.time: units.second,
+    dimension_definitions.current: units.coulomb,
```
blackyblack, line 30: Current is not measured in coulombs.

symplyphysics/core/convert.py
```diff
@@ -17,3 +20,30 @@ def convert_to(value: Expr, target_unit: Expr) -> Expr:
...
+    dimension_definitions.temperature: units.kelvin,
+    dimension_definitions.amount_of_substance: units.mole,
+    dimension_definitions.luminous_intensity: units.candela,
+}
+
+
+def convert_to_si(value: Expr) -> Expr:
+    quantity = value if isinstance(value, Quantity) else Quantity(value)
+    dependencies = dimsys_SI.get_dimensional_dependencies(quantity.dimension)
```
blackyblack, line 39: AFAIK this function returns MLT dimensions. So you only need to process mass, length and time.

symplyphysics/core/convert.py
alesanter: Yes, it should be _charge_.

alesanter: Of course, confused it with charge

symplyphysics/core/convert.py
alesanter: image
Why not, it does support every SI unit.

blackyblack (approved):

PR #904, opened 2024-09-13
Patch symbols 1

alesanter:


symplyphysics/laws/quantities/quantity_is_areal_density_times_area.py
```diff
@@ -0,0 +1,61 @@
...
+law = Eq(extensive_quantity, areal_density * area)
+r"""
+:code:`X = sigma_X * A`
+
+Latex:
+    .. math::
+        X = \sigma_X A
+"""
+
+@validate_input(area_=area)
```
blackyblack, line 50: In similar laws we validated that output dimension corresponds to input.

symplyphysics/quantities/__init__.py
```diff
@@ -78,6 +78,30 @@
...
+Vacuum permittivity, also known as permittivity of free space or the electric constant, is the value of the absolute
+dielectric permittivity of classical vacuum.
+"""
+
+vacuum_permeability = Quantity(units.vacuum_permeability,
+    display_symbol="mu_0",
+    display_latex="\\mu_0")
+"""
+Vacuum permeability, also known as permeability of free space of the magnetic constant, is the value of the absolute
+permeability of classical volume.
```
blackyblack, line 95:
```suggestion
permeability of classical vacuum.
```

symplyphysics/symbols/classical_mechanics.py
```diff
@@ -0,0 +1,96 @@
+"""
+Classical mechanics
+===================
+
+Symbols related to classical mechanics.
+"""
+
+from symplyphysics import units, SymbolNew, angle_type, clone_symbol, dimensionless
+
+force = SymbolNew("F", units.force)
```
blackyblack, line 10: I think force should be in dynamics.

symplyphysics/symbols/classical_mechanics.py
```diff
@@ -0,0 +1,96 @@
...
+"""
+Distance to the origin of the coordinate system.
+"""
+
+distance_to_axis = SymbolNew("r", units.length)
+"""
+Distance to reference axis.
+"""
+
+length = SymbolNew("l", units.length)
```
blackyblack, line 42: This looks quite basic to me. Maybe move it there?

symplyphysics/symbols/classical_mechanics.py
```diff
@@ -0,0 +1,96 @@
...
+**Speed** is the rate of change of the object's position with respect to time. It is the scalar counterpart
+of the velocity vector, which is the rate of change of the object's displacement with respect to time.
+"""
+
+acceleration = SymbolNew("a", units.acceleration)
+"""
+**Acceleration** is the rate of change of the object's velocity with respect to time. It is a vector quantity.
+"""
+
+distance = SymbolNew("d", units.length)
```
blackyblack, line 27: Do we want to get rid of it and use `length` instead?

symplyphysics/symbols/classical_mechanics.py
```diff
@@ -0,0 +1,96 @@
...
+"""
+**Distance** is is a measure of the spatial separation between two points.
+"""
+
+radial_distance = SymbolNew("r", units.length)
+"""
+Distance to the origin of the coordinate system.
+"""
+
+distance_to_axis = SymbolNew("r", units.length)
```
blackyblack, line 37: These two are pretty much the same as radius of curvature. Do we need them?

symplyphysics/symbols/classical_mechanics.py
```diff
@@ -0,0 +1,96 @@
...
+"""
+**Angular distance** is a measure of an angular separation between two points.
+"""
+
+angular_wavenumber = SymbolNew("k", angle_type / units.length)
+"""
+**Angular wavenumber** is the spatial analog of temporal frequency equal to radians per unit length.
+"""
+
+wavelength = SymbolNew("lambda", units.length, display_latex="\\lambda")
```
blackyblack, line 78: Maybe it belongs to `waves`? Same for the symbol above.

symplyphysics/symbols/classical_mechanics.py
```diff
@@ -0,0 +1,96 @@
...
+"""
+**Damping ratio** is a dimensionless measure describing how oscillations in a system decay after a disturbance.
+"""
+
+volume = SymbolNew("V", units.volume)
+"""
+Volume is a measure of regions in three-dimensional space.
+"""
+
+impulse = SymbolNew("J", units.momentum)
```
blackyblack, line 93: It's `dynamics` I guess.

symplyphysics/symbols/classical_mechanics.py
```diff
@@ -0,0 +1,96 @@
...
+"""
+**Angular wavenumber** is the spatial analog of temporal frequency equal to radians per unit length.
+"""
+
+wavelength = SymbolNew("lambda", units.length, display_latex="\\lambda")
+"""
+**Wavelength** or **spatial period** is the distance over which the wave's shape repeats.
+"""
+
+damping_ratio = SymbolNew("zeta", dimensionless, display_latex="\\zeta")
```
blackyblack, line 83: Probably smth like `oscillators` category.

symplyphysics/symbols/electrodynamics.py
```diff
@@ -0,0 +1,142 @@
...
+**Electrical resistivity** is a fundamental specific property of a material that measures its electrical resistance or how
+strongly it resists electric current.
+"""
+
+inductance = SymbolNew("L", units.inductance)
+"""
+**Inductance** is the tendency of an electrical conductor to oppose a change in the electric current flowing through it.
+"""
+
+time_constant = SymbolNew("tau", units.time, display_latex="\\tau")
```
blackyblack, line 133:
```suggestion
electric_time_constant = SymbolNew("tau", units.time, display_latex="\\tau")
```

symplyphysics/core/symbols/symbols.py
```diff
@@ -267,5 +267,22 @@ def clone_symbol(source: SymbolNew,
     return SymbolNew(display_symbol_new, source.dimension, display_latex=display_latex_new, **assumptions)
 
 
+def clone_function(
```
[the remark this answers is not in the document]
alesanter: The idea was to make a helper procedure that would convert a symbol into a function so that we would only need to define symbols and convert them to functions whenever needed. I'm guessing I interpreted the naming in a peculiar way, so maybe `clone_as_function` is better.

symplyphysics/definitions/acceleration_is_speed_derivative.py
```diff
@@ -5,29 +5,37 @@
...
+from symplyphysics import (
+    Quantity,
+    validate_input,
+    validate_output,
+    symbols,
+    clone_function,
+)
 
-acceleration = FunctionNew("a(t)", units.acceleration, display_latex="a")
+acceleration = clone_function(symbols.acceleration, display_symbol="a(t)")
```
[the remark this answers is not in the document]
alesanter: What do you mean?

symplyphysics/laws/astronomy/ratio_of_luminosities_from_ratio_of_masses_of_stars.py
```diff
@@ -23,8 +23,8 @@
 ## L2 - illuminance of second object,
 ## L1 - illuminance of first object.
 
-mass_first = clone_symbol(symbols.basic.mass)
-mass_second = clone_symbol(symbols.basic.mass)
+mass_first = clone_symbol(symbols.mass, display_symbol="m_1")
```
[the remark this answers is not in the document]
alesanter: Isn't it clunky? Moreover it's inconsistent with alphabetical subscripts.

symplyphysics/core/symbols/symbols.py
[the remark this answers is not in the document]
blackyblack: Yes, naming is wrong. `clone_as_function` sounds good.

symplyphysics/definitions/acceleration_is_speed_derivative.py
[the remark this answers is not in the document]
blackyblack: Down below you manually generate latex description.

symplyphysics/laws/astronomy/ratio_of_luminosities_from_ratio_of_masses_of_stars.py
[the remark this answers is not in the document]
blackyblack: It is more consistent with latex representation, where there is no undescore. As for alphabetical subscripts, we do not have other options to separate symbol and subscript.

symplyphysics/laws/electricity/circuits/voltage_across_charging_capacitor_in_serial_resistor_capacitor_circuit.py
```diff
@@ -10,74 +10,61 @@
...
 
 _resistance = time_constant_law.resistance
 _capacitance = time_constant_law.capacitance
 
-_capacitor_current = Function("_capacitor_current", units.current)
-_resistor_current = Function("_resistor_current", units.current)
-_resistor_voltage = Function("_resistor_voltage", units.voltage)
+_capacitor_current = clone_function(symbols.current, display_symbol="I_C")
+_resistor_current = clone_function(symbols.current, display_symbol="I_R")
+_resistor_voltage = clone_function(symbols.voltage, display_symbol="V_R")
```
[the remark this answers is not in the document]
alesanter: To me, the point of making a `symbols` registry is also to remove the need to create symbols of physical quantities from scratch entirely, including derivations. On the contrary, it's clearer and does not interfere with calculations or performance in any way.

symplyphysics/symbols/electrodynamics.py
alesanter: I haven't found any mentions of time constants in other domains of physics though.

symplyphysics/laws/quantities/quantity_is_areal_density_times_area.py
alesanter: I now think this is not needed. Multiplying anything by area would always result in a quantity whose dimension is equal to multiplying that dimension by the dimension of area. Perhaps if we included a dimension to check for as a parameter, this would be useful to have. But so far this check does nothing.

blackyblack: It's true. But we also can show the output dimension this way. For example, `validate_output` never triggers, if formula is correct. But it shows the resulting dimension to the user.

symplyphysics/symbols/electrodynamics.py
blackyblack: Yes, but it sounds too general this way.

symplyphysics/symbols/classical_mechanics.py
alesanter: Operationally they're different. The distance to axis is just the radial distance in cylindrical coordinates and does not imply any motion. Radius of curvature is related to the trajectiory of an object.

symplyphysics/symbols/classical_mechanics.py
alesanter: I rewrote the description of length so that length would refer to the size of an object.

symplyphysics/symbols/classical_mechanics.py
blackyblack: Let's equate such symbols instead of cloning: `distance_to_axis = radius_of_curvature`

alesanter: What if one wants to use both of them in one law? For this reason I believe all symbols here should be distinct.

My point is that if two quantities are operationally different, they should be described by different symbols. So, for instance, distance (magnitude of the displacement vector), length (a dimension), distance to origin, distance to axis, radius of curvature, optical path length, proper length, etc all require distinct symbols because they describe different measurements.

symplyphysics/symbols/electrodynamics.py
alesanter: Okay

symplyphysics/symbols/classical_mechanics.py
blackyblack: It seems to be quite a tricky problem. Let's keep it the way you did. Maybe eventually we'll find a way to consistently define global symbols.

symplyphysics/laws/electricity/electric_field_outside_charged_sphere.py
```diff
@@ -17,44 +17,33 @@
...
 Distance to the center of the sphere.
-
-Symbol:
-    :code:`r`
 """
 
-law = Eq(electric_field_strength, charge / (4 * pi * units.vacuum_permittivity * distance**2))
-r"""
-:code:`E = 1 / (4 * pi * epsilon_0) * q / r^2`
+law = Eq(electric_field_strength, charge / (4 * pi * quantities.vacuum_permittivity * distance**2))
```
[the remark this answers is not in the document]
alesanter: I added a pre-processed role for that in the header.

I see what you suggest, but that makes me think that `vacuum_permittivity` is a variable quantity, which it isn't.

symplyphysics/laws/quantities/quantity_is_areal_density_times_area.py
alesanter: Hm, I see what you mean.

symplyphysics/symbols/classical_mechanics.py
alesanter: I suggest we don't overengineer now since we can re-order them later without changing anything in the documentation or modules.

alesanter: So in my opinion this whole separation of symbols into several files must be 'an implementation detail' and we shouldn't worry too much about what goes where.

blackyblack: Ok. Eventually I want to move them to let users find them easily.

symplyphysics/laws/astronomy/ratio_of_luminosities_from_ratio_of_masses_of_stars.py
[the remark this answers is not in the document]
alesanter: I'm not sure that Latex is the main thing here. To me it would be more consistent to have, say `q_1 + q_2 = q_total` than `q1 + q2 = q_total`. Plus, the underscore is a clear indication that it corresponds to a subscript in Latex and is not part of the name or something.

symplyphysics/core/symbols/symbols.py
```diff
@@ -251,24 +251,25 @@ def print_expression(expr: Expr | Equality | Sequence[Expr | Equality]) -> str:
...
         assumptions) == 0 else assumptions
-    display_symbol_new = source.display_name if display_symbol is None else display_symbol
-    display_latex_new = display_latex
-    if display_latex_new is None:
-        display_latex_new = display_symbol
-    if display_latex_new is None:
-        display_latex_new = source.display_latex
-    return SymbolNew(display_symbol_new, source.dimension, display_latex=display_latex_new, **assumptions)
+    display_symbol = display_symbol or source.display_name
+    display_latex = display_latex or source.display_latex or display_symbol
```
blackyblack, line 261: You can write `display_latex = display_latex or source.display_latex` as `source.display_latex` cannot be `None`.
Same for `clone_function`

symplyphysics/laws/astronomy/relationship_of_periods_of_planets_with_mass_and_semi_axes_of_orbits.py
```diff
@@ -27,8 +27,8 @@
 second_period = Symbol("second_period", units.time)
 first_semi_axis = Symbol("first_semi_axis", units.length)
 second_semi_axis = Symbol("second_semi_axis", units.length)
-first_mass = symbols.mass
-second_mass = symbols.mass
+first_mass = clone_symbol(symbols.mass, "m_1")
```
blackyblack: Latex symbol is unset.

symplyphysics/laws/electricity/electrostatic_force_via_charges_and_distance.py
```diff
@@ -22,17 +22,17 @@
...
 second_charge = clone_as_symbol(symbols.charge, display_symbol="q_2")
 """
-Second charge.
+Second :symbols:`charge`.
 """
 
 distance = symbols.distance_to_origin
 """
-Distance between the charges.
+:symbols:`distance` between the charges.
```
blackyblack, line 35: `distance_to_origin`
IMO it should be `length` here.

symplyphysics/laws/electricity/electrostatic_potential_due_to_point_charge.py
```diff
@@ -16,22 +16,22 @@
...
 absolute_permittivity = symbols.absolute_permittivity
 """
-Absolute permittivity of the medium.
+:symbols:`absolute_permittivity` of the medium.
 """
 
 distance = symbols.distance_to_origin
 """
-Distance to the point charge.
+:symbols:`distance` to the point charge.
```
blackyblack, line 29: `distance_to_origin`

symplyphysics/quantities/__init__.py
```diff
@@ -101,6 +101,34 @@
 or, equivalently, the magnitude of the negative charge carried by a single electron.
 """
 
+hbar = Quantity(units.hbar, display_symbol="hbar", display_latex="\\hbar")
+"""
+**Reduced Planck constnat** is a modified version of the Planck constant used in the description of
```
blackyblack, line 106: constant

plots/thermodynamics/carnot_cycle.py
```diff
@@ -101,6 +110,9 @@
     adiabatic_law.final_volume: carnot_cycle_volume
 })
 
+# print(result_pressure_isothermal_expansion)
+# import sys; sys.exit(1)
```
blackyblack, line 114: Remove

symplyphysics/laws/astronomy/ratio_of_luminosities_from_ratio_of_masses_of_stars.py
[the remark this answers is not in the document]
blackyblack: I tend to use Pythonic style even for physical formulas. In programming we name variables in snake style, i.e. `some_variable` but do not separate numbers with undescores, i.e. `some_variable1`.

symplyphysics/symbols/classical_mechanics.py
alesanter: Well I made it so that the user would, say, write `symbols.length` using the common namespace and not `symbols.classical_mechanics.length`.

symplyphysics/laws/electricity/electrostatic_force_via_charges_and_distance.py
alesanter: Length is defined as the size of some object and not really the distance between two charges.

The problem I faced here is that `distance` is shown as `d` while I want it to be `r`. Therefore I opted for using `symbols.distance_to_origin` in place of `symbols.distance` for the actual symbol so that it would be shown as `r`, but used `` :symbols:`distance` `` in the description. And if I write `` :symbols:`distance_to_origin` `` the final text would read `distance_to_origin between the charges`, which is not what we want, I guess.

symplyphysics/symbols/classical_mechanics.py
blackyblack: Common namespace for symbols is a separate issue.

symplyphysics/laws/electricity/electrostatic_force_via_charges_and_distance.py
blackyblack: It's a bad idea. We could use description like this: `Distance between the charges. See :symbols:distance_to_origin`. Another option is to use `d` in formula. It look more logical to have `d` for distance between charges, than `r`

blackyblack (approved):

symplyphysics/laws/electricity/electrostatic_force_via_charges_and_distance.py
alesanter: Fixed

blackyblack (approved):

PR #907, opened 2024-09-20
Patch symbols

alesanter:


symplyphysics/definitions/sound_level_in_decibels.py
```diff
@@ -9,49 +9,37 @@
...
 """
-The intensity of a sound wave, relative to which the sound level is measured.
+The intensity of a sound wave relative to which the sound level is measured.
 """
 
 definition = Eq(sound_level, reference_sound_level * log(intensity / reference_intensity, 10))
-r"""
-:code:`beta = beta0 * log10(I / I0)`
+"""
+:laws:symbol::
```
blackyblack, line 40: Logarithms with not natural base are not properly rendered. They are shown as `ln(x) / ln(10)`. I suggest to manually write formula here.
Also add explanatory comment every time we stick to manual formula description.

symplyphysics/definitions/thermodynamic_compressibility.py
```diff
@@ -16,49 +16,34 @@
...
 definition = Eq(compressibility, -1 * Derivative(volume(pressure), pressure) / volume(pressure))
-r"""
-:code:`beta = -1 / V(p) * Derivative(V(p), p)`
+"""
+:laws:symbol::
 
-Latex:
-    .. math::
-        \beta = - \frac{1}{V(p)} \frac{\partial V}{\partial p}
+:laws:latex::
```
blackyblack, line 46: Partials are not rendered properly. Some other parameters should exist in function to let latex generator know, that it is a partial derivative. Use manual formula description.

symplyphysics/definitions/volumetric_coefficient_of_thermal_expansion.py
```diff
@@ -50,12 +44,10 @@
...
 )
-r"""
-:code:`alpha_V = 1 / V(T, p) * Derivative(V(T, p), T)`
+"""
+:laws:symbol::
 
-Lambda:
-    .. math::
-        \alpha_V = \frac{1}{V(T, p)} \left( \frac{\partial V}{\partial T} \right)_p
+:laws:latex::
```
blackyblack, line 50: Same. Also `p` suffix is non-standard. There is no way to autogenerate it like this.

symplyphysics/definitions/wave_equation_in_one_dimension.py
```diff
@@ -38,48 +40,37 @@
...
 """
 
 definition = Eq(
     Derivative(displacement(position, time), position, 2),
     Derivative(displacement(position, time), time, 2) / phase_speed**2,
 )
-r"""
-:code:`Derivative(u(x, t), (x, 2)) = (1/v^2) * Derivative(u(x, t), (t, 2))`
+"""
+:laws:symbol::
```
blackyblack, line 63: `displacement` is not `SymbolNew`. Autogeneration may fail to properly process it. Use manual description.

symplyphysics/symbols/classical_mechanics.py
```diff
@@ -47,6 +52,12 @@
 **Length** is a measure of a size of an object.
 """
 
+thickness = SymbolNew("d", units.length)
+"""
+**Thickness** is a measure of a size of an object, usually the separation between two layers, or
```
blackyblack, line 57: It's pretty much the description of `distance`.

alesanter: The question is are they interchangeable? "The distance of the box is 10 cm" and "the thickness of the two points is 3 km" do not make sense.

symplyphysics/definitions/wave_equation_in_one_dimension.py
alesanter: You're right, I forgot

symplyphysics/symbols/classical_mechanics.py
blackyblack: Ok. What about `length`? Length, width and height are pretty much interchangeable.

alesanter: We can have laws that feature all three parameters at once. [Here](https://github.com/blackyblack/symplyphysics/blob/master/symplyphysics/laws/kinematics/rotational_inertia/geometries/slab_about_perpendicular_axis_through_center.py) I imagine we can and need to use both `length` and `width` since we need to describe two dimensions of the body. We can require the user to clone and rename them as needed but it seems reasonable to me to let them all be here. Almost always `length` is used to talk about the general and biggest dimension of the object, and if it's already in use, `width` is used as the secondary and smaller dimension. `height` is most commonly the dimension along the `z`-axis. `thickness`, on the other hand, is specifically about the separation between layers or similar things, which I mentioned in the text.

blackyblack: I think `thickness` could be understood in two ways:
1. Size of a layer of something, i.e. thickness of asphalt.
2. Separation of two layers of something.

1 is interchangeable with `width` or `height`, depending on position, vertical or horizontal.
2 is interchangeable with `distance` but requires rephrasing, i.e. `thickness of air gap` = `distance between walls of the gap`

I suggest to avoid using 2nd meaning of the thickness and use `distance` instead (with a bit of rephrasing). And use `width` , `height` or `length` whenever we use 1st meaning of the thickness (I doubt we have any).

blackyblack (approved):

symplyphysics/definitions/wave_equation_in_one_dimension.py
```diff
@@ -59,10 +59,12 @@
     Derivative(displacement(position, time), position, 2),
     Derivative(displacement(position, time), time, 2) / phase_speed**2,
 )
-"""
-:laws:symbol::
+r"""
+:code:`Derivative(u(x, t), (x, 2)) = (1/v^2) * Derivative(u(x, t), (t, 2))`
```
blackyblack: A comment should be alaways added, when description is not auto-generated.
BTW I will add `any` dimension, to allow using our symbols instead of SymPy.

PR #914, opened 2024-10-10
Core: arguments in functions

alesanter:


symplyphysics/definitions/damped_harmonic_oscillator_equation.py
```diff
@@ -17,14 +17,14 @@
...
 
-displacement = clone_as_function(symbols.distance, display_symbol="x(t)", display_latex="x")
+time = clone_as_symbol(symbols.time, positive=True)
 """
-Displacement of the oscillating body as a function of time. See :symbols:`distance`.
+:symbols:`time`.
 """
 
-time = clone_as_symbol(symbols.time, positive=True)
+displacement = clone_as_function(symbols.distance, [time])
```
blackyblack, line 25: You probably also want to rename it to `x`

symplyphysics/definitions/linear_coefficient_of_thermal_expansion.py
```diff
@@ -35,27 +35,31 @@
...
-    Derivative(length(temperature), temperature) / length(temperature),
+    Derivative(length(temperature, pressure), temperature) / length(temperature, pressure),
 )
 r"""
 :laws:symbol::
 
 .. only:: comment
 
-    Derivatives with one parameter held constant are not very pretty when autogenerated.
+    Derivatives with parameters held constant are not very pretty when autogenerated.
```
blackyblack, line 62: Actually you can try to autogenerate it. It should look like a partial derivative.

symplyphysics/definitions/radiant_exitance_is_radiant_flux_emitted_per_unit_area.py
```diff
@@ -23,14 +23,14 @@
...
 """
-:symbols:`radiant_flux` emitted from the surface.
+:symbols:`area` of the surface.
 """
 
-area = symbols.area
+radiant_flux = clone_as_function(symbols.radiant_flux, [area])
 """
-:symbols:`area` of the surface.
+:symbols:`radiant_flux` emitted from the surface.
```
blackyblack, line 33: Mention function.

symplyphysics/definitions/thermodynamic_compressibility.py
```diff
@@ -29,14 +29,14 @@
...
 """
-:symbols:`volume` of the gas as a function of pressure and other parameters.
+:symbols:`pressure` in the gas.
 """
 
-pressure = symbols.pressure
+volume = clone_as_function(symbols.volume, [pressure])
 """
-:symbols:`pressure` in the gas.
+:symbols:`volume` of the gas as a function of pressure and other parameters.
```
blackyblack, line 39: You can artificially add other parameters and let autogenerate the result.

symplyphysics/docs/printer_latex.py
```diff
@@ -143,4 +143,9 @@ def _print_log(self, expr: Any) -> str:
 
 
 def latex_str(expr: Any, **settings: Any) -> str:
-    return SymbolLatexPrinter(settings).doprint(expr)
+    printer = SymbolLatexPrinter(settings)
+
+    if isinstance(expr, FunctionNew):
+        expr = expr(*expr.arguments)
```
blackyblack, line 149: It looks weird.

symplyphysics/definitions/wave_equation_in_one_dimension.py
```diff
@@ -26,16 +26,16 @@
...
+position = symbols.position
 """
-Factor representing a displacement from rest position, which could be
-pressure, position, electric field, etc., as a function of position
-and time.
+:symbols:`position`, or spatial variable.
 """
 
-position = symbols.position
+displacement = FunctionNew("u", [position], any_dimension)
```
blackyblack, line 34: Maybe add `time` to arguments?

symplyphysics/definitions/thermodynamic_compressibility.py
alesanter: Can you elaborate?

symplyphysics/docs/printer_latex.py
alesanter: This piece of code means that if a standalone function (without applied arguments) is given as an argument, apply its "default" arguments.

Looking at the implementation of the `Printer._print` function, it seems to always just `__repr__` any function without applied arguments, so we can't just add a `_print_FunctionNew` method since it won't work for the reasons stated, and I decided to put this into `latex_str` (and now in `code_str`).

symplyphysics/definitions/linear_coefficient_of_thermal_expansion.py
alesanter: I want to try to make a prettier version of a derivative with fixed arguments, namely $`\left( \frac{\partial f}{\partial x} \right)_{y}`$, at least for now it doesn't look too difficult to me. If that turns out to be too daring a task, we can resort to the usual printer.

symplyphysics/definitions/thermodynamic_compressibility.py
blackyblack: Smth like `parameters = SymbolNew("p", any_dimension)`

symplyphysics/definitions/linear_coefficient_of_thermal_expansion.py
blackyblack: Latex adds parameters to function so it should look like `df(x, y) / dx`. I think it's enough and we do not need to additionally specify constant parameter.

blackyblack (approved):

blackyblack (approved):

symplyphysics/definitions/thermodynamic_compressibility.py
```diff
@@ -34,22 +36,29 @@
...
 .. only:: comment
 
     The derivative is partial since more parameters are needed to evaluate it properly
 
-Latex:
-    .. math::
-        \beta = - \frac{1}{V(p)} \frac{\partial V}{\partial p}
+    Latex:
+        .. math::
+            \beta = - \frac{1}{V(p)} \frac{\partial V}{\partial p}
```
blackyblack, line 59: I think you can remove this.

alesanter: Of course, I forgot. Will do

blackyblack (approved):

PR #918, opened 2024-11-05
Core: symbolic wrappers

alesanter:


symplyphysics/core/symbols/symbols.py
```diff
@@ -27,23 +27,24 @@ def display_name(self) -> str:
         return self._display_name
 
 
-class DimensionSymbolNew:
+class HasDimension:
```
blackyblack, line 30: I don't like this approach. You can simply use a Dimension inside your class. HasDimension only introduces `dimension` property which looks too small for a separate class.

symplyphysics/core/symbols/symbols.py
```diff
@@ -27,23 +27,24 @@ def display_name(self) -> str:
...
         self._dimension = dimension
-        self._display_name = display_name
-        self._display_latex = display_latex or self._display_name
 
     @property
     def dimension(self) -> Dimension:
         return self._dimension
+    
+
+class HasDisplay:
```
blackyblack, line 41: This looks better but I don't like naming. Class should be implemented in an object, so it should describe class of objects. I suggest something like `DisplayWithLatex`

symplyphysics/core/symbols/symbols.py
```diff
@@ -54,7 +55,19 @@ def display_latex(self) -> str:
         return self._display_latex
 
     def _sympystr(self, p: Printer) -> str:
-        return p.doprint(self.display_name)
+        return p.doprint(self.display_name)    
+
+
+class DimensionSymbolNew(HasDimension, HasDisplay):
```
blackyblack, line 61: This is not needed since you've extracted all logic somewhere. Now it's a wrapper that doesn't do anything.

symplyphysics/core/symbols/wrappers.py
```diff
@@ -0,0 +1,55 @@
+from typing import Self, Any
+from sympy import Symbol as SymSymbol
+from .symbols import HasDimension, next_name
+
+
+class Wrapper(HasDimension, SymSymbol):
```
blackyblack, line 6: Too generic name. I'm thinking about `Operation` but how is it different from `Function`? Maybe try to use existing `Function` implementation to cover `Average` functionality?

alesanter: I think there are some common points between this and `Function` (e.g. they accept an argument, although functions can accept more than one unlike these `Wrappers`), but they're different in that a `Function` always wraps its argument(s) in parentheses in both plain code and LaTeX, and in several frequent cases we need something with a custom "wrapping" functionality. Within plain code though they do look like functions (`avg(K)`, `Delta(T)`, `d(S)`, etc), but in LaTeX that's not always the case ($`\langle K \rangle`$ vs $`\Delta T`$ and $`dS`$).

I agree this functionality can be put in the `operations` folder.

symplyphysics/core/symbols/symbols.py
alesanter: I see what you mean. I guess my concern was mostly connected to the type system, I thought that [this line](https://github.com/blackyblack/symplyphysics/blob/3d19f392d740f648961085483d3466e411c3981d/symplyphysics/core/quantity_decorator.py#L19) could be reduced (after getting rid of old `DimensionSymbol`) to basically `ScalarValue | SymQuantity | HasDimension` and the next line to `Dimension | HasDimension`, since it's logical that any objects that has some dimension information attached to it could be used as an argument to those functions, and thus we would not need to update these type aliases every time we implement some new class or decide to re-structure the class. And I think this is what is done in `sympy` to some extent.

Originally, I just wanted `Wrapper` to be a subclass of `DimensionSymbolNew` (since `Wrapper`s are semantically closer to `Symbol`s and they carry some dimension information), but that would mean I would have to set the `_display_name` and `_display_latex` fields, which is sort of weird since that is implemented in `symplyphysics/docs/printer_*.py` modules.

symplyphysics/core/symbols/wrappers.py
blackyblack: I think we can subclass from `Function` but override printing methods to allow custom display.

alesanter: Okay, I'll give it a try.

symplyphysics/core/symbols/symbols.py
```diff
@@ -27,24 +27,25 @@ def display_name(self) -> str:
         return self._display_name
 
 
-class DimensionSymbolNew:
+class DimensionSymbolNew:  # pylint: disable=too-few-public-methods
```
blackyblack, line 30: This class now looks rather useless. Did you try to remove it and use `Dimension` instead?

symplyphysics/core/operations/single_argument_operation.py
```diff
@@ -0,0 +1,21 @@
...
+class SingleArgumentOperation(DimensionSymbolNew, SymSymbol):  # pylint: disable=too-many-ancestors
+    """
+    This class is for dimensional symbols that act as an operation with a
+    single argument. Its subclasses are intended to be used in custom
+    code printers. The argument is placed at position `0` of the `args`
+    property.
+    """
+
+    def __new__(cls, _argument: Basic, **assumptions: Any) -> Self:
+        return SymSymbol.__new__(cls, next_name("SYS"), **assumptions)
```
blackyblack, line 16: Use `SYM`. I think `SYS` is for coordinate systems.

symplyphysics/core/operations/single_argument_operation.py
```diff
@@ -0,0 +1,21 @@
...
+
+class SingleArgumentOperation(DimensionSymbolNew, SymSymbol):  # pylint: disable=too-many-ancestors
+    """
+    This class is for dimensional symbols that act as an operation with a
+    single argument. Its subclasses are intended to be used in custom
+    code printers. The argument is placed at position `0` of the `args`
+    property.
+    """
+
+    def __new__(cls, _argument: Basic, **assumptions: Any) -> Self:
```
blackyblack, line 15: I think `argument` should be `Expr`.

blackyblack (approved):

symplyphysics/core/symbols/symbols.py
alesanter: But inheriting would imply an `is a` relationship but `DimensionSymbolNew` is not a `Dimension`, it just has a corresponding field.

symplyphysics/core/operations/single_argument_operation.py
alesanter: You're right

symplyphysics/core/symbols/symbols.py
blackyblack: But it only has on field, which is Dimension. So basically, **it is** a Dimension.

alesanter:
Closed in favor of #964

PR #919, opened 2024-11-07
Patch documentation

alesanter:


symplyphysics/symbols/basic.py
```diff
@@ -7,9 +7,14 @@
 
 from sympy.physics import units
 from sympy.physics.units.definitions.dimension_definitions import angle as angle_type
-from symplyphysics.core.dimensions import dimensionless
+from symplyphysics.core.dimensions import dimensionless, any_dimension as _any_dimension
 from symplyphysics.core.symbols.symbols import SymbolNew
 
+any_dimension = SymbolNew("X", _any_dimension)
```
blackyblack, line 13: Wrong name. I suggest `any_quantity` or `anything`.

symplyphysics/symbols/basic.py
```diff
@@ -64,6 +69,9 @@
 radius_of_curvature = SymbolNew("r", units.length)
 """
 **Radius of curvature** is the inverse of curvature and is equal to the distance to the center of curvature.
+
+..
+    TODO are this and distance_to_axis interchangeable?
```
blackyblack, line 74: Guess so :)

symplyphysics/symbols/classical_mechanics.py
```diff
@@ -75,6 +75,11 @@
 **Area** is the size of a region on a two-dimensional surface.
 """
 
+arc_length = SymbolNew("s", units.length)
```
blackyblack, line 78: Maybe write it as `arc_length = distance` ?

test/docs/printer_code_test.py
```diff
@@ -1,7 +1,54 @@
...
+        "charge",
+        "distance",
+        "vacuum_permittivity",
+    ],
+)
+
+
+@fixture(name="test_args")
+def test_args_fixture() -> Args:
+    mass = SymbolNew("m")
```
blackyblack, line 27: Better take them from symbols.

alesanter: The issue was that I changed the representation of a symbol and it crashed. I'm not sure it's the best thing to go and check all test strings when something in `symbols` gets changed.

symplyphysics/symbols/classical_mechanics.py
alesanter: I think we need to distinguish between distance as the length of the smallest path connecting two points, i.e. the length of the segment of a straight line connecting them or equivalently the Euclidean distance, and distance as the overall length of the path of a body during its travel. These are two very different definitions of distance from a physical standpoint, but unfortunately we tend to conflate the two in the language we use to describe physical events. That's why I'm trying to be clear about what the symbols we use in laws actually represent.

`arc_length` belongs to the second definition of distance I've written above. Specifically, it's the length of a part of a curve, which is the same as the Euclidean distance iff the curve is a straight line, so they're not equal in general.

symplyphysics/symbols/basic.py
alesanter: I guess I would leave `radius_of_curvature` untouched and rather replace the symbol in the laws I was concerned about.

symplyphysics/symbols/basic.py
alesanter: You're right, didn't think of that then.

symplyphysics/symbols/classical_mechanics.py
blackyblack: Add small note to the comments, so we won't try to fix it later.

symplyphysics/laws/kinematics/displacement_in_simple_harmonic_motion.py
```diff
@@ -7,80 +7,62 @@
...
+    clone_as_symbol,
+    clone_as_function,
 )
 from symplyphysics.core.expr_comparisons import expr_equals
 from symplyphysics.core.quantity_decorator import validate_output_same
 from symplyphysics.core.symbols.quantities import scale_factor
 from symplyphysics.definitions import harmonic_oscillator_is_second_derivative_equation as harmonic_eqn
 
-displacement = symbols("displacement", real=True)
+displacement = clone_as_symbol(
```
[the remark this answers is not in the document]
alesanter: Will do

symplyphysics/laws/kinematics/rotational_inertia/geometries/slab_about_perpendicular_axis_through_center.py
```diff
@@ -11,40 +11,30 @@
...
 """
-Rotational inertia of the slab.
-
-Symbol:
-    :code:`I`
+:symbols:`rotational_inertia` of the slab.
 """
 
-length = Symbol("length", units.length)
+length = clone_as_symbol(symbols.length, subscript="1")
```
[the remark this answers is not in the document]
alesanter: My idea was that `a` and `b` are often reserved for special parameters in models (vide equations of state). Plus, we only need to name two `length`s that are just two sizes of the body measured in a specific way, so I thought just numbering them with subscripts is justified.

symplyphysics/laws/kinematics/classical_addition_of_velocities.py
```diff
@@ -16,50 +16,36 @@
...
     Quantity,
-    Symbol,
     validate_input,
     validate_output,
+    symbols,
+    clone_as_symbol,
 )
 
-body_speed_in_first_frame = Symbol("body_speed_in_first_frame", units.velocity)
+body_speed_in_first_frame = clone_as_symbol(symbols.speed, subscript="A")
```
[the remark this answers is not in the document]
alesanter: I agree with you.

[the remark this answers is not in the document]
alesanter: No, it's okay. I only use `\text{...}` if it's a name/noun or part of it or an abbreviation, not for actual mathematical single-character symbols.

symplyphysics/laws/kinematics/displacement_in_simple_harmonic_motion.py
```diff
@@ -7,80 +7,62 @@
...
-amplitude = symbols("amplitude", positive=True)
-r"""
-The maximum absolute value of the displacement.
-
-Symbol:
-    :code:`q_max`
-
-Latex:
-    :math:`q_\text{max}`
+amplitude = clone_as_symbol(
```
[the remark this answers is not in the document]
alesanter: I guess I would for now refrain from making it a separate operation. It's actually not just `max(q)` but $`\max_{t \in \mathbb{R}}{|q(t)|}`$ since it's the amplitude and not simply the maximum value of the function.

symplyphysics/laws/electricity/electromotive_force_induced_in_uniformly_rotating_coil.py
```diff
@@ -22,74 +22,54 @@
 #. The angular speed of the coil's rotation constant.
```
blackyblack, line 22: What's the difference between this law and sinusoidally rotating coil? Descriptions are almost identical and formulas are the same.

symplyphysics/laws/kinematics/displacement_in_simple_harmonic_motion.py
```diff
@@ -7,80 +7,62 @@
...
+    clone_as_function,
 )
 from symplyphysics.core.expr_comparisons import expr_equals
 from symplyphysics.core.quantity_decorator import validate_output_same
 from symplyphysics.core.symbols.quantities import scale_factor
 from symplyphysics.definitions import harmonic_oscillator_is_second_derivative_equation as harmonic_eqn
 
-displacement = symbols("displacement", real=True)
+displacement = clone_as_symbol(
+    symbols.any_dimension,
```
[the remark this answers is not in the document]
alesanter: My idea was to avoid the raw usage of the `SymbolNew` constructor (and the like) so that one would only need, in the ideal case, the `symbols` module and the `clone_as` family of functions. This would  In this case, `any_quantity` plays the role of a sort of a wildcard quantity that can represent any other one.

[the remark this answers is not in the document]
blackyblack: Cloning any_quantity is ok.

symplyphysics/laws/electricity/electromotive_force_induced_in_uniformly_rotating_coil.py
alesanter: They're the same. I'll remove one.

PR #943, opened 2024-12-27
Law: free energy change in dielectrics

alesanter:


symplyphysics/laws/thermodynamics/dielectrics/free_energy_change_via_temperature_change_and_electric_field_change.py
```diff
@@ -0,0 +1,98 @@
+"""
+Free energy change via temperature change and electric displacement change
```
blackyblack, line 2: Avoid adding new laws and focus on updating documentation on existing laws.

blackyblack (approved):

symplyphysics/laws/thermodynamics/dielectrics/free_energy_change_via_temperature_change_and_electric_field_change.py
alesanter: Yes, I'll get back to it, I just wanted to give myself a little break from that.

symplyphysics/laws/thermodynamics/dielectrics/free_energy_change_via_temperature_change_and_electric_field_change.py
```diff
@@ -0,0 +1,98 @@
...
+    symbols,
+    units,
+    Quantity,
+    SymbolNew,
+    clone_as_symbol,
+    validate_input,
+    validate_output,
+)
+
+free_energy_density_change = SymbolNew("G", units.energy / units.volume)
```
blackyblack, line 31: Why the same symbol name as for Gibbs energy? Shouldn't be smth like `dH`?

blackyblack (approved):

PR #945, opened 2025-01-10
Relativistic documentation

alesanter:


examples/relativistic/distance_traveled_by_unstable_particle.py
```diff
@@ -12,11 +12,13 @@
 ## till its decay in the laboratory frame of reference, where its lifetime is `20 ns`.
 
 values = {
-    relativistic_time_law.moving_observer_time: Quantity(10 * units.nanosecond),
+    relativistic_time_law.proper_time: Quantity(10 * units.nanosecond),
     relativistic_time_law.relativistic_time: Quantity(20 * units.nanosecond),
 }
 
-particle_speed_expr = solve(relativistic_time_law.law, relativistic_time_law.velocity)[1]
+# for some reason sympy doesn't want to extract the speed out of the square root
```
blackyblack, line 19: Maybe we should make it rational and positive to let sympy extract it.

symplyphysics/laws/relativistic/lorentz_transformation_of_time.py
```diff
@@ -1,54 +1,79 @@
...
-time_in_proper_frame = Symbol("time_in_proper_frame", units.time)
-time_in_lab_frame = Symbol("time_in_lab_frame", units.time)
-position_in_lab_frame = Symbol("position_in_lab_frame", units.length)
-proper_frame_speed_in_lab_frame = Symbol("proper_frame_speed_in_lab_frame", units.velocity)
+time_in_proper_frame = clone_as_symbol(symbols.time, display_symbol="t'", display_latex="t'")
+"""
+:symbols:`time` in proper frame :math:`S'`.
+"""
+
+time_in_lab_frame = symbols.time
```
blackyblack, line 54: It might be a good idea to make naming consistent across laws, i.e. use `position_in_second_frame`, `position_in_first_frame` and so on.

symplyphysics/laws/relativistic/relativistic_kinetic_energy.py
```diff
@@ -1,33 +1,53 @@
+"""
+Relativistic kinetic energy
+===========================
+
+The kinetic energy of an object is the form of energy that is possesses due to its motion.
```
blackyblack, line 5:
```suggestion
The kinetic energy of an object is the form of energy that it possesses due to its motion.
```

symplyphysics/laws/relativistic/relativistic_time_dilation.py
```diff
@@ -1,35 +1,53 @@
+"""
+Relativistic time dilation
+==========================
+
+Time dilation is the difference in elapsed time as measured by two clocks,
+either due to a relative speed between them (special relativity),
+or a difference in gravitational potential between their locations.
```
blackyblack, line 7: In this law general relativity case is not covered. I suggest to rename the law to reflect that it measures time dilation from speed only.

symplyphysics/laws/relativistic/relativistic_time_dilation.py
```diff
@@ -1,35 +1,53 @@
...
-# t is time that has passed as measured by the traveling observer,
-# v is velocity,
-# c is speed of light.
+speed = symbols.speed
+"""
+:symbols:`speed` of the object.
+"""
 
-# Links: Wikipedia, formula in box <https://en.wikipedia.org/wiki/Time_dilation#Simple_inference>
+relativistic_time = symbols.time
```
blackyblack, line 32: I think it is a good idea to add separate symbol for relativistic time. The description for special and general relativity cases should go there.

symplyphysics/laws/relativistic/spacetime_interval.py
```diff
@@ -1,41 +1,66 @@
...
+
+**Notes:**
+
+#. If :math:`\\Delta s**2 > 0`, the spacetime interval is said to be *timelike*. Events with a timelike
+   separation can be causally connected, i.e. one can find an inertial reference frame in which both
+   events happen at the same place at different times.
+#. If :math:`\\Delta s**2 = 0`, the spacetime interval is said to be *lightlike*. Events with a lightlike
+   separation are exactly far enough from each other that light could be present at both events, and they
+   are causally connected.
+#. If :math:`\\Delta s**2 < 0`, the spacetime interval is said to be _spacelike_. Events with a spacelike
```
blackyblack, line 21: Why different formatting for `spacelike`?

symplyphysics/laws/relativistic/relativistic_time_dilation.py
alesanter: I was unsure if we need separate symbols for them all. Proper time and proper length are measured in the rest frame, and it seems reasonable to have separate symbols for them. "Relativistic" time isn't some other kind of time, same for length, they are just measured in the lab frame. Although maybe it's reasonable to have them separate too.

blackyblack: Maybe eventually we will use just time, length and mass without `relativistic` prefix. But for now let's keep it.

plots/relativistic/photon_sail.py
```diff
@@ -120,18 +122,18 @@
 print("Definition of reduced energy of incident wave:\n")
 print(print_expression(reduced_incident_wave_energy_eqn), end="\n\n\n")
 
-reduced_mirror_speed_expr = mirror_speed_expr / speed_of_light
+# reduced_mirror_speed_expr_ = mirror_speed_expr / speed_of_light
```
blackyblack, line 125: Delete

blackyblack (approved):

PR #961, opened 2025-02-07
Matrix code printers

alesanter:


symplyphysics/docs/printer_code.py
```diff
@@ -229,6 +229,20 @@ def _print_Add(self, expr: Expr, _order: bool = False) -> str:
 
         return tex
 
+    def _print_MutableDenseMatrix(self, expr: Matrix) -> str:
+        rows, cols = expr.shape
+
+        if cols == 1:
+            parts = [self._print(elem) for elem in expr]
```
blackyblack, line 236: We might want to do the same for matrix with a single row. So essentially matrix with single row will look the same as with single column. Also matrix with single column will look as transposed.

alesanter: I agree, but in that case I would add `.T` at the end of the row vector. This is similar to how they work in sympy.

blackyblack: Not sure it's really needed cause it's obvious that one of the matrices is transponded. But you can do it if you wish.

alesanter: I can think of some edge cases like `dot([a, b], [c, d])` being ambiguous, so yes, I'd rather it were explicit.

PR #964, opened 2025-02-07
Symbolic wrappers

alesanter:


symplyphysics/core/operations/symbolic.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    This class is intended to be subclassed for custom code printing.
+    """
+
+    factor: Expr
+    """
+    Argument of the wrapper.
+    """
+
+    dimension: Dimension
```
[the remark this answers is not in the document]
alesanter: In that case whenever we would need to get the dimension of this "symbol" we would need to write several lines of code with several imports, instead we can pre-calculate it and store it. What's more `collect_factor_and_dimension` can throw an error, and it's reasonable to catch it early.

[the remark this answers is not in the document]
alesanter: By the way, I tried to alter the code of `collect_factor_and_dimension` so that it could be used for symbolic expressions as well, but I see it now means that `Quantity` would allow symbols (as well as derivatives and functions with "dimensionful" arguments), which is unwanted behavior, I guess. Maybe it's worth turning it into a class that can be instantiated with several options (e.g. `allow_derivative`, `allow_dimensionful_arguments`, `allow_symbols`) and then we could have two separate instances of them?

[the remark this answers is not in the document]
blackyblack: I'm not sure. I could tell by looking at the code. But for now it seems it isn't worth the effort.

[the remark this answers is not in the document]
alesanter: The calculate-dimension function can throw an error due to erroneous expressions, and catching it early is worth the caching. And it also breaks the logic of symbols (SymbolNew, FunctionNew) and symbol-like expressions (like these) that they have a dimension attached to them.

symplyphysics/core/operations/symbolic.py
```diff
@@ -0,0 +1,118 @@
...
+        self.wrap_code = wrap_code
+        self.wrap_latex = wrap_latex
+
+
+class Average(Symbolic):  # pylint: disable=too-many-ancestors
+    """
+    Represents the average value of the argument :math:`x`.
+
+    Code:
+        :code:`avg(x)`
```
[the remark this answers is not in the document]
alesanter: I guess this will bind the code printing with the actual code too much. And I'm not sure this will be useful at all.

symplyphysics/core/operations/symbolic.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Code:
+        :code:`Delta(x)`
+
+    Latex:
+        :math:`\\Delta x`
+    """
+
+
+class ExactDifferential(Symbolic):  # pylint: disable=too-many-ancestors
```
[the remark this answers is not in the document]
alesanter: Well, I initially intended this module to be a sort of a semantic markup for symbols that could be then used in code printing.

But talking about `Derivative`, they're not exactly interchangeable. A derivative requires a function or an expression as the first argument, but in domains like thermodynamics we often have equations that cannot be easily expressed by derivatives and have to rely on differentials. Perhaps one can argue that $`dU = T \, dS - p \, dV`$ is equivalent to the combination of $`\left(\frac{\partial U}{\partial S}\right)_V = T`$ and $`\left(\frac{\partial U}{\partial V}\right)_S = -p`$ but the first is preferred in most cases. And we can always find a way to convert the former representation to the latter.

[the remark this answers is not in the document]
blackyblack: So is it correct to use `Derivative` when we have function as a parameter and `ExactDifferential` when we have a symbol?

symplyphysics/core/operations/symbolic.py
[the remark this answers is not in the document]
blackyblack: It is a common approach in OOP where object is responsible for it's own printing. However SymPy printing does not follow this approach so there is no sense in implementing this.

symplyphysics/docs/printer_code.py
```diff
@@ -245,9 +245,13 @@ def print_row(row: int) -> str:
         return "[" + ", ".join(parts) +"]"
 
     def _print_Average(self, expr: Any) -> str:
+        # The `wrap_code` field is ignored due to being unused.
```
blackyblack, line 248: Move these notes to the related class. Noone will look into printing code to find out about this issue.

symplyphysics/core/dimensions.py
```diff
@@ -132,15 +126,96 @@ def _unsupported_derivative(expr: Derivative) -> tuple[Basic, Dimension]:
...
-    if hasattr(expr, "dimension") and not isinstance(expr, SymQuantity):
-        return expr, getattr(expr, "dimension")
-
     for k, v in cases.items():
         if isinstance(expr, k):
             return v(expr)
     return (expr, dimensionless)
 
 
+def collect_dimension(expr: Expr) -> Dimension:
```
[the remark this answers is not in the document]
alesanter: I see, but these are realizations of two different algorithms. The `scale_factor` of a `Quantity` must be a number and we disallow any free symbols within it. Whereas when we need to find the dimension of a possibly symbolic expression, we don't need that restriction anymore.

Also the notion of scale factor is now ill-formed, e.g. if the expression is `Quantity(2 * units.meter) + symbols.radius`, we can't just return `2 + symbols.radius` as the scale factor since `symbols.radius` is still dimensionful.

And since these `collect` functions are recursive, we can't have a unified implementation for these two algorithms.

symplyphysics/core/operations/symbolic.py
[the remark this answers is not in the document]
alesanter: I guess so, but I don't see why we couldn't use `ExactDifferential` with functions, too. It all depends on what you want to do, differentiating and applying the differential are two a bit different operations. Say, we have an equation $`f(x, y) = h(x) + g(y)`$. Differentiating both sides by `x` would result in $`\frac{\partial}{\partial x} f(x, y) = h'(x)`$. While taking the differential of both sides would result in $`df(x, y) = h'(x) dx + g'(y) dy`$. Sometimes the latter is needed.

symplyphysics/conditions/electricity/maxwell_equations/divergence_of_magnetic_induction_field_is_zero.py
```diff
@@ -17,4 +17,4 @@
 def magnetic_field_divergence_condition(magnetic_induction_: VectorField) -> bool:
     divergence_magnetic_induction = divergence_operator(magnetic_induction_)
     divergence_magnetic_induction_quantity = Quantity(divergence_magnetic_induction)
-    return divergence_magnetic_induction_quantity.scale_factor == 0
+    return bool(divergence_magnetic_induction_quantity.scale_factor == 0)
```
blackyblack, line 20: Why this bool? The result of comaprison should always be a bool.

symplyphysics/core/dimensions.py
```diff
@@ -1,22 +1,21 @@
...
+from sympy.functions.elementary.miscellaneous import MinMaxBase
 from sympy.physics.units import Dimension, Quantity as SymQuantity
-from sympy.physics.units.systems.si import SI
+from sympy.physics.units.prefixes import Prefix
+from sympy.physics.units.systems.si import dimsys_SI
 
 from .errors import UnitsError
 
-ScalarValue: TypeAlias = Expr | float
+ScalarValue: TypeAlias = Expr | float | int
```
blackyblack, line 12: int is a subset of float. I think this is not really needed.

symplyphysics/core/symbols/quantities.py
```diff
@@ -45,32 +51,35 @@ def __init__(self,
...
+    def _eval_Abs(self) -> Quantity:
         return self.__class__(Abs(self.scale_factor), dimension=self.dimension)
 
 
 # Allows for some SymPy comparisons, eg Piecewise function
-@dispatch(Quantity, Quantity)
+@dispatch(Quantity, Quantity)  # type: ignore[misc]
 def _eval_is_ge(lhs: Quantity, rhs: Quantity) -> bool:
-    return lhs.scale_factor >= rhs.scale_factor
+    return bool(lhs.scale_factor >= rhs.scale_factor)
```
blackyblack, line 74: This bool is redundant.

symplyphysics/core/symbols/symbols.py
```diff
@@ -188,7 +192,7 @@ def print_expression(expr: Expr | Equality | Sequence[Expr | Equality]) -> str:
     use_unicode = pprinter.is_unicode()
     uflag = pretty_use_unicode(use_unicode)
     try:
-        return pprinter.doprint(expr)
+        return str(pprinter.doprint(expr))
```
blackyblack, line 195: This should be already str.

symplyphysics/laws/quantum_mechanics/harmonic_oscillator/energy_levels.py
```diff
@@ -21,6 +21,7 @@
 """
 
 from sympy import Eq, Rational, solve, symbols as sym_symbols, sqrt
+from sympy.functions.special.polynomials import hermite as _hermite
```
blackyblack, line 24: You can import with renaming.

test/chemistry/electronic_current_in_probe_circuit_in_gas_plasma_test.py
```diff
@@ -37,7 +37,7 @@ def test_basic_current(test_args: Args) -> None:
         test_args.floating_plasma_potential,
         test_args.probe_potential,
     )
-    assert_equal(result, 425.84 * prefixes.milli * units.ampere)
+    assert_equal(result, 425.84 * units.milli * units.ampere)
```
blackyblack, line 40: Why did you change to units?

alesanter: I added support for the `sympy.physics.units.prefixes.Prefix` class, so I wanted to test it. This also probably means that our `symplyphysics.prefixes` module isn't needed anymore.

symplyphysics/core/dimensions.py
alesanter: Mathematically yes, but Python's `int` is not a subclass of `float`, so for the type checker they aren't interchangeable. I guess I'll remove it anyway for now.

symplyphysics/conditions/electricity/maxwell_equations/divergence_of_magnetic_induction_field_is_zero.py
```diff
@@ -17,4 +17,4 @@
 def magnetic_field_divergence_condition(magnetic_induction_: VectorField) -> bool:
     divergence_magnetic_induction = divergence_operator(magnetic_induction_)
     divergence_magnetic_induction_quantity = Quantity(divergence_magnetic_induction)
-    return bool(divergence_magnetic_induction_quantity.scale_factor == 0)
+    return divergence_magnetic_induction_quantity.scale_factor == 0  # type: ignore[no-any-return]
```
blackyblack, line 20: It's weird. Why does mypy think it's Any type instead of bool? Maybe Python treats this as a SymPy Eq expression?
In this case I suggest to convert scale_factor to float before comparison.

symplyphysics/core/symbols/quantities.py
```diff
@@ -71,7 +71,7 @@ def _eval_Abs(self) -> Quantity:
 # Allows for some SymPy comparisons, eg Piecewise function
 @dispatch(Quantity, Quantity)  # type: ignore[misc]
 def _eval_is_ge(lhs: Quantity, rhs: Quantity) -> bool:
-    return bool(lhs.scale_factor >= rhs.scale_factor)
+    return lhs.scale_factor >= rhs.scale_factor  # type: ignore[no-any-return]
```
blackyblack, line 74: Don't silence the mypy checker. I think it hints that scale factor should be evaluated to a number before comparison.

symplyphysics/core/symbols/symbols.py
```diff
@@ -37,7 +37,7 @@ def display_latex(self) -> str:
         return self._display_latex
 
     def _sympystr(self, p: Printer) -> str:
-        return str(p.doprint(self.display_name))
+        return p.doprint(self.display_name)  # type: ignore[no-any-return]
```
blackyblack, line 40: It seems Printer does not have type hints, that's why it assumes Any here. So here you can forcefully convert to str()

PR #971, opened 2025-03-04
Core: vector symbols and expressions

alesanter:


symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+
+    3. Distributivity of scalar multiplication w.r.t. vector addition: for all scalars `k` and
+       vectors `a, b`, `k * (a + b) = k * a + k * b`.
+
+    The last property, distributivity of scalar multiplication w.r.t. field addition, is not
+    represented within the functionality of `VectorScale`.
+
+    As a consequence of the properties of the vector field, one has
+
+    1. For all vectors `a`, `0 * k = 0`.
```
blackyblack, line 231: Where is a?

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+
+        return VectorScale(vector, scale, evaluate=False)
+
+    def _sympystr(self, p: Printer) -> str:
+        vector, value = self.args
+
+        if isinstance(value, (Add, Mul)):
+            return f"{p.doprint(vector)}*({p.doprint(value)})"
+
+        return f"{p.doprint(vector)}*{p.doprint(value)}"
```
blackyblack, line 295: Is it the same as in the check above?

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+
+    def _eval_nseries(self, x: Any, n: Any, logx: Any, cdir: Any) -> Any:
+        pass
+
+
+class VectorAdd(VectorExpr):
+    """
+    Class representing the notion of vector addition as a property of vectors.
+
+    Note that only
```
blackyblack, line 325: Only what?

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+        pass
+
+
+class VectorAdd(VectorExpr):
+    """
+    Class representing the notion of vector addition as a property of vectors.
+
+    Note that only 
+
+    This operation has the following properties:
```
blackyblack, line 327: No properties.

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+
+            for addend in addends:
+                if isinstance(addend, VectorScale):
+                    mapping[addend.args[0]] += addend.args[1]
+                else:
+                    mapping[addend] += S.One
+
+            return mapping
+
+        def filter_scales(mapping: dict[VectorExpr, Expr]) -> dict[VectorExpr, Expr]:
```
blackyblack, line 379: Not sure it is really needed. Zero scale should result to zero vector that does not affect addition.

test/core/experimental/vectors/test_vector.py
```diff
@@ -0,0 +1,215 @@
...
+    one_newton_force = VectorSymbol(name, dim, norm=1 * units.newton, display_latex=latex)
+    assert_equal(one_newton_force.norm, 1 * units.newton)
+    assert not one_newton_force.is_zero
+
+    # NOTE: Norm of a symbol can't be 0 (probably needs to be fixed)
+    with raises(ValueError):
+        VectorSymbol("a", norm=0)
+
+    # Correct dimension of norm
+    VectorSymbol("S", units.area, norm=1 * units.meter**2)
```
blackyblack, line 41: Do we really need expressions as norm? Something simpler, like Quantity could be easier to support.

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+        if isinstance(value, (Add, Mul)):
+            return f"{p.doprint(vector)}*({p.doprint(value)})"
+
+        return f"{p.doprint(vector)}*{p.doprint(value)}"
+
+    def _eval_evalf(self, prec: int) -> VectorScale:
+        return VectorScale(self.vector.evalf(n=prec), self.scale.evalf(n=prec)).doit()
+
+
+class Scale(Expr):  # type: ignore[misc]
```
blackyblack, line 301: I do not see why this is needed.

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+
+    @property
+    def norm(self) -> Optional[Expr]:
+        return self._norm
+
+    def _hashable_content(self) -> tuple[Any, ...]:
+        return (self._id,)
+
+
+class VectorNorm(Expr):  # type: ignore[misc]
```
blackyblack, line 154: Why do we need this class? I do not see it being used in expressions.

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+
+    The vector argument is stored in position `0` of `self.args`.
+
+    The vector norm has the following properties:
+
+    1. **Subadditivity**: for all vectors `a` and `b`, `norm(a + b) <= norm(a) + norm(b)`.
+
+    1. **Absolute homogeneity**: for all scalars `k` and vectors `a`, `norm(k * a) = abs(k) * norm(a)`.
+
+    1. **Positive definiteness**: for all vectors `a`, `norm(a) = 0` iff `a = 0`.
```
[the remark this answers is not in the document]
alesanter: _iff_ is shorthand for _if and only if_ but now I guess not everyone knows it outside of the English-speaking world

symplyphysics/core/experimental/vectors/__init__.py
alesanter: This is a common mathematical operation on vectors. We'll need it to solve vector equations later.

Or you think it could be replaced by, say `VectorDot` product of the vector with itself? Though in that case writing `sqrt(dot(a, a))` is clunkier than `norm(a)`.

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+
+        return self
+
+    def _eval_nseries(self, x: Any, n: Any, logx: Any, cdir: Any) -> Any:
+        pass
+
+    def _sympystr(self, p: Printer) -> str:
+        return f"norm({p.doprint(self.argument)})"
+
+    def _eval_evalf(self, prec: int) -> VectorNorm:
```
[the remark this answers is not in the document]
alesanter: Hm, I thought initially that the purpose of `.evalf` was to evaluate all numbers within the sub-expressions of the given expression up to a certain number of digits. Say, `(pi * x).evalf(3)` produces `3.14 * x`. But `sin(pi * x).evalf(3)` doesn't do anything. In this case, I'll get rid of the mixin altogether.

symplyphysics/core/experimental/vectors/__init__.py
blackyblack: Nevermind. Keep it as is.

symplyphysics/core/experimental/vectors/__init__.py
alesanter: No, the scale isn't wrapped in parentheses.

symplyphysics/core/experimental/vectors/__init__.py
alesanter: Yes, but we still need to kick those zeroes out of the list of addends. The resulting list will be saved in `._args` and we don't need any `ZERO`es there.

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -0,0 +1,430 @@
...
+        )
+
+        if norm is not None:
+            norm = sympify(norm)
+
+            if not isinstance(norm, Expr):
+                raise TypeError(f"Norm {norm} must be an Expr, got {type(norm).__name__}.")
+
+            try:
+                is_non_negative = bool(norm >= 0)
```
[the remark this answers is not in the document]
alesanter: I opted in for `ask(Q.negative(...))` in the end. Is `.is_negative` defined on all expressions?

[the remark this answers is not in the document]
blackyblack: No idea. Probably not.

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -172,7 +168,7 @@ class VectorNorm(Expr):  # type: ignore[misc]
     def argument(self) -> VectorExpr:
         return self.args[0]  # type: ignore[no-any-return]
 
-    # NOTE: Add __new__ that would dispatch the code execusing depending on the value of `vector`
+    # NOTE: Add __new__ that would dispatch the code execusion depending on the value of `vector`
```
blackyblack, line 180: execution

test/core/experimental/vectors/test_vector.py
alesanter: Technically we already use symbols instead of norms, say in $`F = m a`$. So from this standpoint it doesn't seem too big a stretch to me to allow symbols as norms.

Also the norm of a vector symbol $`\mathbf{F}`$ would appear as $`F`$ in all expressions right away, not as $`\mathrm{norm}(\mathbf{F})`$, and we wouldn't need to `.subs` the latter later.

blackyblack: Still not sure it is a good idea but let it be.

PR #972, opened 2025-03-05
Patch core functionality

alesanter:


test/core/vectors/arithmetics_test.py
```diff
@@ -577,7 +577,7 @@ def check(v: Vector, correct: list[ScalarValue]) -> None:
 def test_cartesian_vector_rejection() -> None:
     w = Vector([1, 2, -1])
 
-    def check(v: Vector, correct: list[ScalarValue]) -> None:
+    def check(v: Vector, correct: list[Any]) -> None:
```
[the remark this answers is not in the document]
alesanter: I see. I think `SupportsFloat` encompasses what we need, although maybe it needs a separate type alias (because technically `Expr` is subclass of `SupportsFloat` but not all `Expr`s actually support float). Also, I didn't replace it in the old vector and field modules since they're kind of wrong anyway and we don't want to support them.

symplyphysics/core/symbols/quantities.py
```diff
@@ -97,12 +98,20 @@ def _eval_is_ge(lhs: Quantity, rhs: Quantity) -> bool:
...
+    subs_: dict[Expr, SymQuantity],
+) -> Sequence[Quantity]:
     return [Quantity(sympify(c).subs(subs_)) for c in input_]
 
 
-def scale_factor(quantity_: Quantity | float) -> float:
-    if isinstance(quantity_, Quantity):
+def scale_factor(quantity_: SupportsFloat) -> float:
+    """
+    Exctracts the scale factor converted to `float` from the input if it is a quantity. Otherwise
```
blackyblack, line 110: Extracts

test/core/experimental/vectors/test_vector.py
```diff
@@ -35,7 +35,8 @@ def test_init() -> None:
         VectorSymbol(name, dim, norm=symbols.force, display_latex=latex)
 
     one_newton_force = VectorSymbol(name, dim, norm=1 * units.newton, display_latex=latex)
-    assert_equal(one_newton_force.norm, 1 * units.newton)
+    if one_newton_force.norm:  # mypy complains otherwise
```
blackyblack, line 38: It's bad idea to silence mypy like this. Now you can skip the assertion check in test if norm is None or false. Use `norm()` or assert norm is not None instead of `if`

blackyblack (approved):

PR #975, opened 2025-03-07
Scalar conversions between coordinate systems

alesanter:


symplyphysics/core/experimental/coordinate_systems/convert_base_scalars.py
```diff
@@ -0,0 +1,133 @@
...
+    return {
+        r: sqrt(rho**2 + z**2),
+        theta: atan2(rho, z),
+        phi: phi_,
+    }
+
+
+@dispatch(BaseCoordinateSystem, BaseCoordinateSystem)  # type: ignore[no-redef]
+# pylint: disable-next=function-redefined
+def convert_base_scalars(
```
blackyblack, line 113: Is this really needed? Looks weird.

test/core/experimental/coordinate_systems/test_scalar_conversion.py
```diff
@@ -0,0 +1,161 @@
...
+
+def apply_conversion(point: Point, conversion: ScalarConversion) -> Point:
+    return {new_coordinate: expr.subs(point) for (new_coordinate, expr) in conversion.items()}
+
+
+def equal_points(point1: Point, point2: Point) -> bool:
+    return all(expr_equals(coordinate1, point2[s]) for (s, coordinate1) in point1.items())
+
+
+def test_cartesian_to_cartesian(test_args: Args) -> None:
```
blackyblack, line 49: Is it possible that base scalar x in new coordinate system is 2x in old coordinate system?

blackyblack, line 49: Add negative test, when points do not match.

symplyphysics/core/experimental/coordinate_systems/convert_base_scalars.py
alesanter: I don't think we should disallow conversions between same types of coordinate systems. I'll add a comment.

test/core/experimental/coordinate_systems/test_scalar_conversion.py
alesanter: A coordinate system is defined up to the Lamé coefficients, and in this case they will be `(2, 2, 2)`, so we would need to instantiate a new coordinate system class. This would suffice:

```python
class ScaledCartesianCoordinateSystem(CartesianCoordinateSystem):

    @property
    def lame_coefficients(self) -> tuple[Expr, Expr, Expr]:
        return S(2), S(2), S(2)
```

alesanter: And then we would define the conversions:

```python
@dispatch(CartesianCoordinateSystem, ScaledCartesianCoordinateSystem)
def convert_base_scalars(
    old_system: CartesianCoordinateSystem,
    new_system: ScaledCartesianCoordinateSystem,
) -> ScalarConversion:
    x, y, z = old_system.base_scalars
    x1, y1, z1 = old_system.base_scalars
    return {
        x: x1 / 2,
        y: y1 / 2,
        z: z1 / 2,
    }


@dispatch(ScaledCartesianCoordinateSystem, CartesianCoordinateSystem)
def convert_base_scalars(
    old_system: CartesianCoordinateSystem,
    new_system: ScaledCartesianCoordinateSystem,
) -> ScalarConversion:
    x1, y1, z1 = old_system.base_scalars
    x, y, z = new_system.base_scalars
    return {
        x1: 2 * x,
        y1: 2 * y,
        z1: 2 * z,
    }
```

symplyphysics/core/experimental/coordinate_systems/convert_base_scalars.py
alesanter: Also for the option of the missing conversion, we could later add a way to try to compose it using existing coordinate systems. For instance if the conversions `A -> B` and `B -> C` (and `B` is most often the Cartesian coordinate system) are defined but `A -> C` isn't, we can compose it from those two.

PR #978, opened 2025-03-13
Solve vector equations

alesanter:


symplyphysics/core/experimental/solvers/__init__.py
```diff
@@ -0,0 +1,89 @@
+from typing import Callable, Any
+from sympy import Basic, S, Eq, sympify, solve as sym_solve
+from ..vectors import VectorExpr, VectorSymbol, norm, VectorAdd
+
+
+def apply(eqn: Any, f: Callable[[Basic], Basic], *, zero: Basic = S.Zero) -> Eq:
+    """
+    Applies `f` to both sides of the equation `eqn`, if it is an equality, else treats `f` as the
```
blackyblack, line 8:
```suggestion
    Applies `f` to both sides of the equation `eqn`, if it is an equality, else treats `eqn` as the
```

symplyphysics/core/experimental/solvers/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+    combination = expr.as_symbol_combination()
+    i = None
+
+    for j, (v, _) in enumerate(combination):
+        if vector_equals(v, atomic):
+            i = j
+            break
+
+    if i is None:
+        raise ValueError("...")
```
blackyblack, line 79: Maybe write something here?

symplyphysics/laws/dynamics/vector/acceleration_from_force.py
```diff
@@ -36,28 +40,44 @@ def force_law(acceleration_: Vector) -> Vector:
     return scale_vector(mass, acceleration_)
 
 
+# NOTE: Proof of concept for the experimental features
```
blackyblack, line 43: Either rewrite it as a test, or rewrite the law to use new feature.

symplyphysics/laws/dynamics/vector/acceleration_from_force.py
```diff
@@ -36,28 +40,44 @@ def force_law(acceleration_: Vector) -> Vector:
...
+__acceleration = VectorSymbol("a", units.acceleration)
+
+__force_law = Eq(__force, __acceleration * mass)
+
+__acceleration_law = express_atomic(__force_law, __acceleration)
+assert vector_equals(__acceleration_law.lhs, __acceleration)
+assert vector_equals(__acceleration_law.rhs, __force / mass)
+
+__mass_eqn = apply(__force_law, norm)
+__mass_law = solve_into_eq(__mass_eqn, mass)[0]
```
blackyblack, line 55: It's quite confusing, how to choose, if you should use `express_atomic` or `solve_into_eq` depending on the symbol. Is it possible to implement some general `solve()` method that selects the proper method automatically?

test/core/experimental/solvers/test_solve.py
```diff
@@ -0,0 +1,84 @@
...
+    c = VectorSymbol("c")
+
+    with evaluate(False):
+        lhs = a + b
+        rhs = b + a
+    assert vector_equals(lhs, rhs)
+
+    with evaluate(False):
+        lhs = a + c
+        rhs = c + b + a - b  # FIXME: Recursion error when evaluating scale factor
```
blackyblack, line 52: This PR won't be approved before this is fixed.

test/core/experimental/solvers/test_solve.py
```diff
@@ -0,0 +1,84 @@
...
+    assert vector_equals(lhs, rhs)
+
+    with evaluate(False):
+        lhs = a + b * 3 - a * 2 + a - b - b - b  # FIXME: Recursion error when evaluating scale factor
+    assert vector_equals(lhs, ZERO)
+
+
+def test_express_atomic() -> None:
```
blackyblack, line 60: Add test with `reduce_factor`

symplyphysics/laws/dynamics/vector/acceleration_from_force.py
alesanter: Though we can write a function that would dispatch on the type of the symbol we're trying to solve for, it would only be helpful in a very limited number of situations. And even then we would still need to derive other relations ourselves. This is partly because all other vector operations are irreversible (such as `dot`, `cross`, and `mixed`), and because it's not always evident what strategy should be used to solve the equation. For example, with this equation ($`\mathbf{F} = m \mathbf{a}`$) to solve for $`m`$ we can't just call `solve` (or `solve_into_eq`) on it, we have to apply `norm` on both sides first. But if we take another common equation, like $\mathbf{r} = \mathbf{r}_0 + \mathbf{v}_0 t + \frac{1}{2} \mathbf{a} t^2$, and attempt to solve for $`t`$, `norm` wouldn't help, and it's actually not obvious what to do here unless more constraints are given.

blackyblack: Ok, but imagine a typical user of SymPy library. His general approach is to use `solve()` on the equation. But for VectorExpr it doesn't work and he's got no idea, where to look at. Ideally we would like to have `solve_for_vector()` but I understand that it's very hard to implement properly. So I want to make `express_atomic()` more visible and more obvious that it is an entry point to solve vector equations. I suggest to think about better naming and elaborate on description to point that it's a replacement for `solve()` function.

PR #979, opened 2025-03-13
Dot and cross product

alesanter:


symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -359,63 +442,468 @@ class VectorAdd(VectorExpr):
...
+
+        for lhs_v, lhs_s in lhs.as_symbol_combination():
+            for rhs_v, rhs_s in rhs.as_symbol_combination():
+                result += cls.from_atomic(lhs_v, rhs_v) * lhs_s * rhs_s
+
+        return result
+
+    @classmethod
+    def from_symbols(cls, lhs: VectorSymbol, rhs: VectorSymbol) -> Expr:
+        sign, args = sort_with_sign((lhs, rhs), key=_atomic_registry.get)
```
blackyblack, line 579: I don't quite understand how it works. How do you know lhs and rhs symbols are not reordered due to sort_with_sign?

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -359,63 +442,468 @@ class VectorAdd(VectorExpr):
...
+
+
+dot = VectorDot  # pylint: disable=invalid-name
+
+
+class VectorCross(VectorExpr):
+    """
+    The **cross product**, or **vector product**, is a binary operation that takes two vectors and
+    returns another vector. The cross product is only defined in a *3-dimensional space* (although
+    the its construction is also possible in a 7-dimensional space, the following properties do not
```
blackyblack, line 627:
```suggestion
    its construction is also possible in a 7-dimensional space, the following properties do not
```

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -359,63 +442,468 @@ class VectorAdd(VectorExpr):
...
+        return self.from_symbols(self.lhs, self.rhs)
+
+    def as_symbol_combination(self) -> tuple[tuple[AtomicVectorExpr, Expr], ...]:
+        sign, args = sort_with_sign((self.lhs, self.rhs), key=_atomic_registry.get)
+
+        if sign == 0:
+            return ()
+
+        if sign == -1:
+            result = type(self)(*args, evaluate=False)
```
blackyblack, line 802: Could you explain, how it works?

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -359,63 +442,468 @@ class VectorAdd(VectorExpr):
...
+
+            # Refer to formula #2
+            return (  # type: ignore[no-any-return]
+                c * VectorDot.from_atomic(lhs, d) - d * VectorDot.from_atomic(lhs, c))
+
+        # Refer to formula #1
+        return VectorSymbolCross.from_symbols(lhs, rhs)
+
+
+class VectorSymbolCross(VectorCross):
```
blackyblack, line 768: How is it different from just VectorCross? And why there are no similar classes for Dot and  Mixed products?

symplyphysics/core/experimental/vectors/__init__.py
alesanter: That's where `sign` comes into use, and what's more, we don't need to know the exact way the values have been shuffled, only the sign of the permutation. With the dot product, the order doesn't matter, so we only need to know if the two arguments are the same (i.e. `sign == 0`). With cross and mixed product, however, if the given array of elements is an odd number of swaps away from the ordered array (i.e. the permutation of the elements is odd), we have to multiply the result (with the elements ordered) by negative one.

symplyphysics/core/experimental/vectors/__init__.py
alesanter: It is different from `VectorCross` in that we can know for sure that both of its arguments are `VectorSymbol` and not just arbitrary `VectorExpr`. I wanted to make sure that `VectorExpr` could be expressed as a linear combination of irreducible vectors, and with the classes developed so far, it would be just `VectorSymbol` and `VectorSymbolCross` — all other vector expressions can be reduced to a linear combination of them. Dot and mixed product don't have such requirements and therefore don't need to have separate subclasses.

symplyphysics/core/experimental/vectors/__init__.py
alesanter: When `sort_with_sign` sorts `(lhs, rhs)`, it might have swapped them, which would mean we need to construct a new `VectorSymbolCross` object with the arguments in the correct order, but then we would also multiply the end vector by `-1` (this is covered by `S(sign)` at the end). If `lhs` and `rhs` are already in the right order, do nothing and just return the instance itself, and `sign` will be `1` in that case.

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -1,18 +1,68 @@
...
+
+    _symbol_registry: Registry[VectorSymbol]
+    _cross_registry: Registry[VectorSymbolCross]
+
+    def __init__(self) -> None:
+        self._symbol_registry = Registry()
+        self._cross_registry = Registry()
+
+    def add(self, value: AtomicVectorExpr) -> None:
+        if isinstance(value, VectorSymbol):
```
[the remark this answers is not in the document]
alesanter: It already has that type per definition of `AtomicVectorExpr`.

symplyphysics/core/experimental/vectors/__init__.py
```diff
@@ -137,47 +181,54 @@ def __new__(
...
         if not display_symbol:
-            display_symbol = f"VEC{self._id}"
+            display_symbol = f"VEC{id_}"
             if not display_latex:
-                display_latex = f"\\mathbf{{v}}_{{{self._id}}}"
+                display_latex = f"\\mathbf{{v}}_{{{id_}}}"
         elif not display_latex:
             display_latex = f"\\mathbf{{{display_symbol}}}"
 
         DimensionSymbol.__init__(
```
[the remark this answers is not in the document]
alesanter: You're right. Actually in this case half of this function can be safely moved into `__init__`. What I observed is that in some previous versions `__init__` was called several times and I thought I for that reason need to move everything up into `__new`__, but I've checked and so far `__init__` is only called once in this class.

test/core/experimental/vectors/test_vector.py
```diff
@@ -6,10 +6,10 @@
 from symplyphysics.core.experimental.vectors import (
     VectorSymbol,
     ZERO,
-    norm,
+    VectorNorm as norm,
```
blackyblack, line 9: I suggest to instead make helper functions, i.e.
```
def norm(vector: VectorExpr, **kwargs: Any):
    return VectorNorm(vector, kwargs)
```

Now it may be confusing to see an object created when you expected to call a function `norm()`

blackyblack (approved):

PR #985, opened 2025-05-09
Refactor coordinate systems

alesanter:


symplyphysics/core/experimental/coordinate_systems/new_coordinate_systems.py
```diff
@@ -0,0 +1,355 @@
...
+    def lame_coefficients(self, base_scalars: Optional[Sequence[Expr]] = None) -> Sequence[Expr]:
+        """
+        An `i`-th Lamé coefficient is the length of the derivative of the position vector `r` with
+        respect to the `i`-th base scalar of the system in question.
+        """
+
+        subs = self.base_scalar_subs(base_scalars)
+        assumption = self.assumption(base_scalars)
+
+        return tuple(
```
blackyblack, line 165: This code should be in `generate` function. And here you simply return generated coefficients.

symplyphysics/core/experimental/coordinate_systems/new_coordinate_systems.py
```diff
@@ -0,0 +1,355 @@
...
+
+            lame_coefficients = []
+
+            for i_row in range(3):
+                row = obj._cartesian_derivative_matrix.row(i_row)
+                lame_coefficient = sqrt(sum(elem**2 for elem in row)).refine(assumption).simplify()
+                lame_coefficients.append(lame_coefficient)
+
+            h1, h2, h3 = lame_coefficients
+            obj._lame_coefficients = h1, h2, h3
```
blackyblack, line 80: This code looks like an overkill. Inherited class can decide to use this kind of lame coefficients generation. But base class should not try to guess what inherited class will do.

symplyphysics/core/experimental/coordinate_systems/new_coordinate_systems.py
```diff
@@ -0,0 +1,355 @@
...
+
+        for i_row in range(diff_matrix.rows):
+            for i_col in range(diff_matrix.cols):
+                elem = diff_matrix[i_row, i_col]
+
+                def const(expr: Expr, _: object) -> Expr:
+                    return expr
+
+                for func_base_scalar, base_scalar in zip(self._base_scalar_functions, base_scalars):
+                    elem = elem.replace(func_base_scalar, partial(const, base_scalar))
```
blackyblack, line 224: Replacing with `base_scalar` instead of partially applied function does not work?

symplyphysics/core/experimental/coordinate_systems/new_coordinate_systems.py
```diff
@@ -0,0 +1,355 @@
...
+    def generate_lame_coefficients(self) -> tuple[Expr, Expr, Expr]:
+        return S.One, S.One, S.One
+
+    def cartesian_derivative_matrix(
+        self,
+        _base_scalars: Optional[Sequence[Expr]] = None,
+    ) -> Matrix:
+        return Matrix.eye(3)
+
+    def lame_coefficients(
```
blackyblack, line 255: Don't think it should be overriden.

symplyphysics/core/experimental/coordinate_systems/new_coordinate_systems.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def inverse_cartesian_transform(self, cartesian_scalars: Sequence[Expr]) -> Sequence[Expr]:
+        """
+        Defined how base scalars `{q_i}` of the system in question can be expressed using Cartesian
+        scalars `x, y, z`.
+        """
+
+        raise NotImplementedError
+
+    def assumption(self, base_scalars: Optional[Sequence[Expr]] = None) -> Boolean:  # pylint: disable=unused-argument
```
blackyblack, line 129: Do we have an idea, what kind of assumption is this?

symplyphysics/core/experimental/coordinate_systems/new_coordinate_systems.py
alesanter: I wanted to pass a usual closure `lambda _: base_scalar`, but `pylint` argues it's a `cell-var-from-loop` error. But I see now that the tests all pass if I replace the `partial` with that closure. So maybe it's okay to silence `pylint` here then.

symplyphysics/core/experimental/coordinate_systems/new_coordinate_systems.py
alesanter: The actual assumptions differ from system to system. It's more of a help for `sympy` when simplifying expressions, since it doesn't know, for instance, that `sin(theta)` is never negative in a spherical coordinate system. So this has to be provided by each type of coordinate system separately.

symplyphysics/core/experimental/coordinate_systems/new_coordinate_systems.py
alesanter: I will move their generation from `__new__` to `generate_lame_coefficients`, but the user might not want to use the pre-defined `self._base_scalars`. So here they are replaced with that the user has provided (the `base_scalars` parameter).

symplyphysics/core/experimental/coordinate_systems/new_coordinate_systems.py
blackyblack: I mean it is impossible to deduce, what are the assumptions and what they do, from their name.

PR #986, opened 2025-05-09
Add coordinate vector functionality

alesanter:


symplyphysics/core/experimental/vectors/new_coordinate_vectors.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def __str__(self) -> str:
+        name = type(self.system).__name__.removesuffix("CoordinateSystem")
+
+        inner = ", ".join(str(component) for component in self.components)
+
+        return f"{name}Vector({inner})"
+
+    def _eval_vector_norm(self) -> Expr:
+        return sqrt(sum(elem**2 for elem in self.components))
```
blackyblack, line 54: Is it true for any coordinate system?

test/core/experimental/vectors/test_new_coordinate_vectors.py
```diff
@@ -0,0 +1,171 @@
...
+
+    t = test_args.t
+    assert vector_diff(v1, t) == 0
+
+    vt = CoordinateVector([t, t, -t], test_args.cart)
+    assert vector_diff(vt) == CoordinateVector([1, 1, -1], test_args.cart)
+
+
+def test_cylindrical_coordinate_vector(test_args: Args) -> None:
+    v1 = CoordinateVector([1, 1, 1], test_args.cyl)
```
blackyblack, line 62: If it is coordinates in cylindrical coordinate system, its norm is `sqrt(2)` I guess.

symplyphysics/core/experimental/vectors/new_coordinate_vectors.py
alesanter: We're only working with orthogonal coordinate systems which means the base vectors are orthogonal to each other everywhere, so this is true by design.

blackyblack: If we are working only with orthogonal coordinate systems, how did we attach cylindrical coordinate system to this vector?

test/core/experimental/vectors/test_new_coordinate_vectors.py
blackyblack: So what does this vector correspond to?
I mean `CoordinateVector([1, 1, 1], test_args.cyl)` - what is it? Cylindrical coordinate system assumes it is a vector in cylindrical coordinate system. And `[1, 1, 1]` are vector components? In cylindrical coordinate system I expected them to be radius, height and angle.

alesanter: Github didn't render my reply correctly so I `pandoc`ed it into the pdf attached: [reply.pdf](https://github.com/user-attachments/files/20234028/reply.pdf)

symplyphysics/core/experimental/vectors/new_coordinate_vectors.py
alesanter: A cylindrical coordinate system is an orthogonal system because its base vectors are orthogonal to one another at all points in space. The same is true for a spherical system. Basically the base vectors in a non-Cartesian coordinate system can be obtained by a rotation of the usual vectors $`\mathbf{i}, \mathbf{j}, \mathbf{k}`$ so that the right angles between them are conserved.

test/core/experimental/vectors/test_new_coordinate_vectors.py
blackyblack: Ok. But what is `CoordinateVector([1, 1, 1], test_args.cyl)`? What are 1, 1, 1 components? Is it like multiplicators to base vectors?

alesanter: Yes

alesanter: I mean, this is what components of a vector generally are. We have a (local) basis and the coefficients to which the base vectors are multiplied are called the components of the vector.

symplyphysics/core/experimental/vectors/new_coordinate_vectors.py
```diff
@@ -1,15 +1,105 @@
...
+        super().__init__()
+
+        if isinstance(coordinates, Sized):
+            n = len(coordinates)  # can't extract out of if-block, mypy complains otherwise
+        else:
+            coordinates = tuple(coordinates)
+            n = len(coordinates)
+
+        if n != 3:
+            raise ValueError(f"The point must have all 3 coordinates defined, got {n}.")
```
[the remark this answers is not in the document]
alesanter: This would only work in a Cartesian system though. Furthermore, this way only the $z$ coordinate can be $0$. We can instead supply a separate function that would fill in the missing coordinate, but again, this would only work in a Cartesian (or cylindrical) system. 1D is even more restricting on that account.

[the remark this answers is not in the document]
blackyblack: I think we can x, y and z to 0 in this order for cartesian. R, angle phi, angle theta for spherical (make sure that theta is elevation angle). And R, phi, z for cylindrical. So it pretty much works for every coordinate system we support.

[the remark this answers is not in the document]
alesanter: For it to work in a spherical system the polar angle must not be $0$ (or $\pi$) though, otherwise the azimuth is undefined. Also, we can't switch the polar angle and the azimuth, they must come in the order $r, \theta, \varphi$.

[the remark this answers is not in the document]
blackyblack: There are papers where angles are switched to another order. But I see your point here.

PR #1000, opened 2025-06-05
Patch vector definition of momentum

alesanter:


symplyphysics/definitions/vector/momentum_is_mass_times_velocity_vector.py
```diff
@@ -9,94 +9,62 @@
...
-def calculate_velocity(mass_: Quantity, momentum_: QuantityVector) -> QuantityVector:
-    result_vector = velocity_law(momentum_.to_base_vector())
-    return QuantityVector.from_base_vector(result_vector, subs={mass: mass_})
+@validate_input(mass_=mass, velocity_=velocity)
+@validate_output(momentum)
+def calculate_momentum(
+    mass_: Quantity,
+    velocity_: QuantityCoordinateVector,
+) -> Expr:
+    momentum_expr = solve_for_vector(law, momentum).rhs
```
blackyblack, line 49: Fix this function to avoid `rhs` everywhere.

symplyphysics/definitions/vector/momentum_is_mass_times_velocity_vector.py
```diff
@@ -9,94 +9,62 @@
...
+    momentum_value = momentum_expr.subs({
+        mass: mass_,
+        velocity: velocity_,
+    })
+
+    return combine_coordinate_vectors(momentum_value)
```
blackyblack, line 55: I still don't understand this function. Probably should be renamed or avoided completely.

alesanter: The issue is that `sympy` doesn't automatically add up or scale `CoordinateVector`s, so I wrote this function to do that. And here I call it to get a single `QuantityCoordinateVector` as a result, and not an unprocessed multiplication of the mass quantity and the velocity quantity vector.

blackyblack: > I suppose renaming it to simplify_coordinate_vectors would make it clearer.
Yes. But having a single Vector object is easier to understand. Simplification assumes we could still have a combination of expressions and it is not clear, why initial combination of expressions was worse than resulting and why should we `combine` or `simplify` before returning from a function.

symplyphysics/core/experimental/coordinate_systems/vector.py
```diff
@@ -149,6 +149,18 @@ def _eval_derivative(self, symbol: Expr) -> VectorExpr:
 
         return CoordinateVector(diff_components + diff_base_vectors, self.system, self.point)
 
+    @classmethod
+    def from_expr(cls, expr: Expr) -> Expr:
```
blackyblack, line 153: I expect this to return an instance of this class.

alesanter: Okay, the closest would be `Self | sympy.core.numbers.Zero`, since `CoordinateVector([0, 0, 0])` always returns `S.Zero`.

alesanter: Although `Self` would work too, `mypy` doesn't really check that deep.

PR #1001, opened 2025-06-05
Patch vector definition of angular momentum 

alesanter:


symplyphysics/definitions/vector/angular_momentum_is_position_cross_linear_momentum.py
```diff
@@ -17,69 +17,48 @@
...
-
-    :return: pseudovector of angular momentum of the particle.
+position_vector = clone_as_vector_symbol(symbols.distance_to_origin)
+"""
+Displacement of the body relative to the origin of the reference frame. See
+:symbols:`distance_to_origin`.
+"""
 
-        Symbol: :code:`L`
+linear_momentum = clone_as_vector_symbol(symbols.momentum)
```
blackyblack, line 32: Could we use arrow for vector symbols instead of bold font?

blackyblack (approved):

symplyphysics/definitions/vector/angular_momentum_is_position_cross_linear_momentum.py
alesanter: We could, but boldface is used more often in physics, especially for typesetting. And it's less clunky than the over-arrow.

blackyblack: Yes but IMO it's hard to distinguish bold symbol from normal symbol, especially if they are not aside to each other.

PR #1038, opened 2025-06-12
Patch vector law of relativistic momentum

alesanter:


symplyphysics/laws/relativistic/vector/relativistic_momentum.py
```diff
@@ -1,59 +1,83 @@
...
+:laws:symbol::
 
+:laws:latex::
+"""
 
-def velocity_law(momentum_: Vector) -> Vector:
-    p_dot_p = dot_vectors(momentum_, momentum_)
-    factor = speed_of_light / sqrt((rest_mass * speed_of_light)**2 + p_dot_p)
-    return scale_vector(factor, momentum_)
+velocity_law = Eq(
```
blackyblack, line 55: Cannot it be solved from momentum law?

blackyblack (approved):

symplyphysics/laws/relativistic/vector/relativistic_momentum.py
alesanter: Not directly, since velocity appears in the dot product too, so we need to express `dot(v, v)` in terms of `dot(p, p)` first.

PR #1047, opened 2025-06-17
Patch vector law of the curl of electric field

alesanter:


symplyphysics/core/dimensions/dimensions.py
```diff
@@ -82,9 +83,22 @@ def print_dimension(dimension: Dimension) -> str:
     return "dimensionless" if dimsys_SI.is_dimensionless(dimension) else str(dimension.name)
 
 
+def assert_quantity_point(point: AppliedPoint, func: Optional[str] = None) -> None:
```
blackyblack, line 86: Does not look like it belongs here.

blackyblack (approved):

blackyblack (approved):

PR #1066, opened 2025-06-22
Improve descriptions

blackyblack:


symplyphysics/definitions/acceleration_is_speed_derivative.py
```diff
@@ -2,11 +2,16 @@
...
+
+**Conditions:**
+
+#. Speed must be differentiable with respect to time on the interval of interest.
+#. Non-relativistic motion, i.e. speed of a body is much less than the speed of light and space-time curvature is negligible.
 
 **Links:**
 
-#. `Wikipedia <https://en.wikipedia.org/wiki/Acceleration#Instantaneous_acceleration>`__.
+#. `Wikipedia – Instantaneous acceleration <https://en.wikipedia.org/wiki/Acceleration#Instantaneous_acceleration>`__
```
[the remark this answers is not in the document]
blackyblack: I think with article name it looks better in the documentation.

symplyphysics/definitions/lorentz_factor.py
```diff
@@ -2,17 +2,21 @@
...
 
 **Notation:**
 
-#. :quantity_notation:`speed_of_light`.
+#. :quantity_notation:`speed_of_light`
+
+**Conditions:**
+
+#. The speed is non-negative and less than speed of light.
+#. Special-relativistic effects only (flat space-time, no gravity).
```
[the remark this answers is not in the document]
blackyblack: It's hard to properly formulate it so I will keep it for now.

symplyphysics/definitions/compliance_is_inverse_stiffness.py
```diff
@@ -2,12 +2,16 @@
 Compliance is inverse stiffness
 ===============================
 
-*Compliance, or flexibility*, of a spring is the inverse of its stiffness and measures
-how flexible the spring is.
+*Compliance* (also called *flexibility*) quantifies how much a system deforms under load; for an ideal linear spring it is the reciprocal of its stiffness.
+
+**Conditions:**
+
+#. Stiffness is non-zero and constant (linear, time-invariant spring).
```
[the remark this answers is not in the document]
alesanter: We might have there
```
This law holds locally (i.e. at a specific point along the spring), and is always true when stiffness is constant throughout the spring.
```

symplyphysics/definitions/density_from_mass_volume.py
```diff
@@ -2,13 +2,16 @@
...
 
-Volumetric mass *density* of an object is a physical quantity equal to the mass of the
-object per unit of its volume. See :doc:`laws.quantities.quantity_is_volumetric_density_times_volume`
-for a general version of this law.
+Volumetric mass *density* is the mass of substance contained in a unit volume of a substance. See :doc:`laws.quantities.quantity_is_volumetric_density_times_volume`
+for a generalized version of this law.
+
+**Conditions:**
+
+#. The material is homogeneous over the volume considered (uniform density).
```
alesanter, line 10: This law is applicable if we view density as an average density of a given volume of the body if the mass isn't distributed evenly.

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
```diff
@@ -2,13 +2,18 @@
...
 
-Assuming there is a damping force acting on an oscillating body that is linearly proportional
-to the body's velocity, we can write a differential equation for the body's position. We're
-assuming the body only moves in one direction.
+Describes the motion of a single-degree-of-freedom mechanical oscillator that experiences *linear (viscous) damping*.
+
+**Conditions:**
+
+#. Damping force is directly proportional to velocity (viscous).
+#. Undamped angular frequency is positive and constant.
```
alesanter, line 10: Angular frequency cannot be negative, so we can get rid of the `positive` condition.

Also, the equation doesn't require it to be constant. It would only matter for the solution of this equation.

symplyphysics/definitions/damped_harmonic_oscillator_equation.py
```diff
@@ -2,13 +2,18 @@
...
+#. Damping ratio is non-negative.
```
alesanter, line 11: I don't see how this equation requires it to be non-negative. Also, it is almost always non-negative anyway, the system is unstable if `zeta < 0`.

symplyphysics/definitions/compressibility_factor_is_deviation_from_ideal_gas.py
```diff
@@ -1,29 +1,28 @@
...
-the temperature or the larger the pressure.
+The **compressibility factor** (also called the **compression factor** or **gas deviation factor**) quantifies how far a real gas departs from ideal-gas behaviour.
 
 **Notation:**
 
 #. :quantity_notation:`molar_gas_constant`.
 
+**Conditions:**
+
+#. Temperature is positive.
```
alesanter, line 13: We can leave it out since it is always true.

symplyphysics/definitions/angular_wavenumber_is_inverse_wavelength.py
```diff
@@ -2,11 +2,16 @@
 Angular wavenumber is inverse wavelength
 ========================================
 
-*Angular wavenumber* is the spatial frequency of a wave, measured in radians per unit distance.
+*Angular wavenumber* is the spatial frequency of a wave in classical wave theory.
+
+**Conditions:**
+
+#. Wavelength is positive.
```
alesanter, line 9: Wavelength is always positive since it's defined as a positive distance.

PR #1069, opened 2025-06-25
Torque due to electric dipole in uniform electric field

alesanter:


symplyphysics/laws/electricity/vector/torque_due_to_electric_dipole_in_uniform_electric_field.py
```diff
@@ -0,0 +1,65 @@
...
+Torque due to electric dipole moment in uniform electric field
+==============================================================
+
+An object with an electric dipole moment is subject to a torque when placed in an external
+electric field. The torque tends to align the dipole with the field.
+
+**Notes:**
+
+#. The dipole receives *no* overall net force due to the fact that the forces exerted on the point
+   charges constituting the dipole cancel each other.
```
blackyblack, line 11: I doubt it is true, since dipole has non-zero length, more distant end of the dipole has smaller acting force than closer end. And why do we need this note at all?

blackyblack (approved):

symplyphysics/laws/electricity/vector/torque_due_to_electric_dipole_in_uniform_electric_field.py
alesanter: The electrostatic force is calculated as $`\mathbf{F} = q \mathbf{E}`$. The charges have the same magnitude and different signs, the electric field $`\mathbf{E}`$ is constant in the region of the dipole. therefore $`\mathbf{F}_+ + \mathbf{F}_- = |q| \mathbf{E} - |q| \mathbf{E} = 0`$.

I think we can get remove this note, I just thought this is related to the topic of torques and it's not an entirely obvious observation, but I wouldn't put it into a separate law.

blackyblack: Where do you see that electric field is constant? I see the condition that it is uniform, not constant.

alesanter: A uniform field is by definition constant

PR #1100, opened 2025-07-08
Core: surface integral

alesanter:


test/core/experimental/integrals/surface_integral_test.py
```diff
@@ -0,0 +1,137 @@
...
+    assert_equal(result.doit(), 6 * m**2)
+
+    # More complex surface and integrand
+
+    parametrization = AppliedPoint([
+        m * sin(t1) * cos(t2),
+        m * sin(t1) * sin(t2),
+        m * cos(t1),
+    ], CARTESIAN)
+    surface = Surface((t1, t2), parametrization)
```
blackyblack, line 63: Maybe call it `parameters` and `function`?

blackyblack (approved):

test/core/experimental/integrals/surface_integral_test.py
alesanter: Parametrization seems to be a [common name](https://en.wikipedia.org/wiki/Parametrization_(geometry)) for this.

PR #1103, opened 2025-07-09
Derive torque and angular momentum relationship

alesanter:


symplyphysics/laws/dynamics/vector/torque_is_angular_momentum_derivative.py
```diff
@@ -42,6 +54,66 @@
...
+    _newtons_second_law.force,
+    _torque_def.force,
+)
+
+_torque_eqn = _torque_def.law.subs({
+    _torque_def.torque: torque(time),
+    _torque_def.position_vector: _position(time),
+})
+
+_derived_law = apply(
```
blackyblack, line 99: It looks suspicious. How does it know which derivatives to apply to?

symplyphysics/laws/dynamics/vector/torque_is_angular_momentum_derivative.py
```diff
@@ -42,6 +54,66 @@
...
+    _angular_momentum_eqn,
+    lambda x: vector_diff(x, time),
+).subs(
```
blackyblack, line 102: Maybe solving for a system of equations would look better?

symplyphysics/laws/dynamics/vector/torque_is_angular_momentum_derivative.py
alesanter: I'm not sure what you mean by your question. It's generally true that if $`a = b`$, then $`\frac{\partial a}{\partial x} = \frac{\partial b}{\partial x}`$ for any $`x`$.

symplyphysics/laws/dynamics/vector/torque_is_angular_momentum_derivative.py
alesanter: It would be much better, we don't have it for vectors yet. For now I believe the way I did here is logically sound enough.

PR #1107, opened 2025-07-11
Derive buoyant force law

alesanter:


symplyphysics/laws/dynamics/buoyant_force_from_density_and_volume.py
```diff
@@ -39,6 +46,77 @@
...
+    global_index,
+    Idx("i", (1, 2)),
+).doit().subs({
+    _superposition_law.force[1]: _top_force_expr,
+    _superposition_law.force[2]: _bottom_force_expr,
+})
+_net_force_expr = CoordinateVector.from_expr(_net_force_expr)
+
+# The axis along which depth is measure in this law is antiparallel to the z-axis
+_net_force_projection = VectorDot(_net_force_expr, -1 * _k).doit()
```
blackyblack, line 104: Don't quite understand why we project force to -k vector.

symplyphysics/laws/dynamics/buoyant_force_from_density_and_volume.py
```diff
@@ -39,6 +46,77 @@
...
+# Depth is the difference in the `z` coordinates between the given point and the top side
+_depth = _pressure_depth_law.height
+
+_top_depth = clone_as_symbol(_depth, subscript="0")
+_bottom_depth = _top_depth + _z_length
+
+# a.k.a. gauge pressure; it is a function of `_depth`
+_hydrostatic_pressure = _pressure_depth_law.law.rhs.subs(_pressure_depth_law.density, fluid_density)
+
+_atmospheric_pressure = clone_as_symbol(symbols.pressure, subscript="0")
```
blackyblack, line 64: Why do we need this? I think the law works without atmosphere.

symplyphysics/laws/dynamics/buoyant_force_from_density_and_volume.py
alesanter: This justifies the original minus sign in the law. I suppose we can get rid of it since it doesn't bring much information.

symplyphysics/laws/dynamics/buoyant_force_from_density_and_volume.py
```diff
@@ -98,22 +109,23 @@
...
     ],
     (buoyant_force, _z_length),
     dict=True,
 )[0][buoyant_force]
 
-assert expr_equals(law.rhs, _net_force_projection)
+assert expr_equals(law.rhs, _net_force_norm_expr)
 
-# For any other shape of the object, we can approximate it with parallelepipeds
+# For any other shape of the object, refer to a more general law that connects the buoyant force
```
blackyblack, line 126: It is possible to derive this law for any kind of surface. Although the proof is more complex. I think we should either stick to approximation with parallelepipeds or use another proof. This comment about more general law is hardly relevant here.

alesanter: I guess you're right.

blackyblack (approved):

PR #1117, opened 2025-07-13
Derive gravity potential law

alesanter:


symplyphysics/laws/dynamics/potential_energy_from_mass_and_height.py
```diff
@@ -44,6 +56,73 @@
...
+})
+
+# Now let's show that `U = -W`
+
+_k0 = _work_energy_law.kinetic_energy(_work_energy_law.time_before)
+_k1 = _work_energy_law.kinetic_energy(_work_energy_law.time_after)
+
+# The initial value doesn't matter since we only calculate the difference in potential energy:
+_u0 = 0
+_u1 = potential_energy
```
blackyblack, line 97: How do you know it's potential energy, not minus potential energy?

alesanter: The `potential_energy` in this law _is_ the potential energy of the body at the final point of its movement, by definition. I don't see why it could take a negative sign.

blackyblack: No, you've stated this is difference of the body potential energy. If final value is the resulting potential energy of a body, you should justify initial zero energy.

alesanter: Ok, I now made this step clearer.

PR #1175, opened 2025-12-12
Rename laws

alesanter:


test/thermodynamics/equations_of_state/van_der_waals/reduced_van_der_waals_pressure_test.py
```diff
@@ -6,7 +6,7 @@
     units,
     Quantity,
 )
-from symplyphysics.laws.thermodynamics.equations_of_state.van_der_waals import reduced_pressure
+from symplyphysics.laws.thermodynamics.equations_of_state.van_der_waals import reduced_van_der_waals_pressure
```
blackyblack, line 9: You can use `law` for short in all tests.

blackyblack (approved):

test/thermodynamics/equations_of_state/van_der_waals/reduced_van_der_waals_pressure_test.py
alesanter: Yes, it was mostly left from other tests. I automated the process of fixing the imports, so this is why you can see this change.

